PR #9, opened 2016-02-06
Fix a typo: Aspect tatio -> Aspect ratio

mitya57:
Just a minor typo. Noticed while testing that D-Bus menu stuff :)


zaps166:
Thanks :)
This typo was also in 22.10.2012 (the oldest QMPlay2 archive which I have) :smile:

PR #21, opened 2016-06-12
Add CMake Build scripts

arthurzam:
This is quite a big commit consisting of all the CMake script files of all the project (based on QMake's subdir approach).
I tested building with Qt5 on my main Gentoo system and Qt4 & Qt5 on my Arch Linux system. I hope you have some other distributions so we can check all the possible problems (which I think come to Qt finding problem or pkg-config).

As a user of Gentoo, where I like many options to compile with, I decided to add many compile options, making it much easier to customize the build. All those options with short documentation are found in /CMakeLists.txt file.

In every CMakeLists.txt you will find an array of HDR (headers), SRC (sources) and RESOURCES. If you would like to add some files, just append them in a new line in the appropriate array. Also, in the gui project, I added an empty GUI_FORMS array for all the future *.ui files.

I tried to follow your install & build scripts for UNIX, but maybe I missed some of them. Please tell if you find.

For building, run those commands (after you cd to the root directory of this project):

1. mkdir -p build && cd build
2. cmake ..
3. make -j$(nproc)
4. make DESTDIR=$(pwd)/app install

Those commands will compile all the options (with Qt5) and install after the compilation into the app folder.

I suggest that we fix all the complains and only after them we merge. After we merge this PR, I would like to update the Arch's PKGBUILD and Gentoo's ebuild to use the new CMake.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,7 @@
 
 #Project config
 *.pro.user
+CMakeLists.txt.user
+
+#Build Directories
+build/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,57 @@
+project(QMPlay2)
+cmake_minimum_required(VERSION 3.0.2)
+include(GNUInstallDirs)
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    set(DEFAULT_ALSA ON)
+    set(DEFAULT_PORTAUDIO OFF)
+else()
+    set(DEFAULT_ALSA OFF)
+    set(DEFAULT_PORTAUDIO ON)
+endif()
+
+OPTION(USE_QT5 "Build with Qt5" ON)
+OPTION(USE_TAGLIB "Build with tags editor" ON)
+OPTION(USE_AVRESAMPLE "Build with support for libavresample" OFF)
+
+OPTION(USE_ALSA "Build with ALSA support" ${DEFAULT_ALSA})
+OPTION(USE_AUDIOCD "Build with AudioCD support" ON)
+OPTION(USE_OPENGL "Build with OpenGL support" ON)
+OPTION(USE_PORTAUDIO "Build with PortAudio support" ${DEFAULT_PORTAUDIO})
+OPTION(USE_PULSEAUDIO "Build with PulseAudio support" ON)
+OPTION(USE_XVIDEO "Build with XVideo support" ON)
+
+OPTION(USE_CHIPTUNE_GME "Build Chiptune with GME support" ON)
+OPTION(USE_CHIPTUNE_SID "Build Chiptune with SIDPLAY support" ON)
+
+OPTION(USE_FFMPEG_VAAPI "Build VAAPI acceleration into FFmpeg" ON)
+OPTION(USE_FFMPEG_VDPAU "Build VDPAU acceleration into FFmpeg" ON)
+OPTION(USE_FFMPEG_AVDEVICE "Build FFmpeg with libavdevice suport" ON)
+
+set(LANGUAGES "All" CACHE STRING "A space-seperated list of translations to compile in to QMPlay2, or \"None\", or \"All\".")
+
+# set default build as Release
+if(NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE "Release")
+endif()
+
+# get QMPLAY2_VERSION by running the version script
+execute_process(
+    COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/version
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    OUTPUT_VARIABLE QMPLAY2_VERSION)
+string(STRIP ${QMPLAY2_VERSION} QMPLAY2_VERSION)
+
+add_subdirectory(src)
+add_subdirectory(lang)
+
+install(FILES ChangeLog DESTINATION ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATAROOTDIR}/qmplay2/)
+
+# uninstall target
+configure_file(
+    "${CMAKE_CURRENT_SOURCE_DIR}/src/cmake_uninstall.cmake.in"
+    "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake"
+    IMMEDIATE @ONLY)
+
+add_custom_target(uninstall
+    COMMAND ${CMAKE_COMMAND} -P ${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake)
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,7 @@ Changes in QMPlay2 build 16.06.14:
 	- fixed color in graphics subtitles (e.g. dvdsub),
 	- hide mouse cursor when cover image is visible,
 	- updated Russian translation (victorr2007),
+	- added CMake build (Zamarin Arthur),
 	- show more information to the user,
 	- bugfix,
 
--- a/compile_unix
+++ b/compile_unix
@@ -70,8 +70,8 @@ else
 		VERSION=$(./version)
 		if [ $? == 0 ]; then
 			mkdir -p app/share/man/man1
-			cp src/gui/Unix/QMPlay2.1 app/share/man/man1
-			sed -i "s/VERSION/$VERSION/" app/share/man/man1/QMPlay2.1
+			cp src/gui/Unix/QMPlay2.1.in app/share/man/man1/QMPlay2.1
+			sed -i "s/@QMPLAY2_VERSION@/$VERSION/" app/share/man/man1/QMPlay2.1
 			gzip -f app/share/man/man1/QMPlay2.1
 		fi
 	fi
--- a/lang/CMakeLists.txt
+++ b/lang/CMakeLists.txt
@@ -1,0 +1,29 @@
+project(langs)
+cmake_minimum_required(VERSION 3.0.2)
+include(GNUInstallDirs)
+
+if(LANGUAGES STREQUAL "All")
+    # build LANGUAGES from all existing .ts files
+    file(GLOB QMPLAY2_TSS *.ts)
+elseif(NOT LANGUAGES OR LANGUAGES STREQUAL "None")
+    set (QMPLAY2_TSS "")
+else()
+    string(REGEX MATCHALL [a-zA-Z_@]+
+           LANGUAGES_TR ${LANGUAGES})
+    set(QMPLAY2_TSS "")
+    foreach(LANG ${LANGUAGES_TR})
+       LIST(APPEND QMPLAY2_TSS "${LANG}.ts")
+    endforeach()
+endif ()
+
+if(USE_QT5)
+    find_package(Qt5LinguistTools REQUIRED)
+    qt5_add_translation(QM_FILES ${QMPLAY2_TSS})
+else()
+    find_package(Qt4 REQUIRED)
+    qt4_add_translation(QM_FILES ${QMPLAY2_TSS})
+endif()
+
+add_custom_target(translations ALL DEPENDS ${QM_FILES})
+
+install(FILES ${QM_FILES} DESTINATION ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATAROOTDIR}/qmplay2/lang/)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,0 +1,18 @@
+cmake_minimum_required(VERSION 3.0.2)
+
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+set(CMAKE_AUTOMOC ON)
+if(USE_QT5)
+    find_package(Qt5Widgets REQUIRED)
+    if(Qt5Widgets_VERSION LESS 5.6.1)
+        message(AUTHOR_WARNING "Qt5 >= 5.6.1 is recommended for stable usage")
+    endif()
+else()
+    find_package(Qt4 REQUIRED)
+    include("${QT_USE_FILE}")
+    add_definitions(${QT_DEFINITIONS})
+endif()
+
+add_subdirectory(gui)
+add_subdirectory(qmplay2)
+add_subdirectory(modules)
--- a/src/cmake_uninstall.cmake.in
+++ b/src/cmake_uninstall.cmake.in
@@ -1,0 +1,21 @@
+if(NOT EXISTS "@CMAKE_CURRENT_BINARY_DIR@/install_manifest.txt")
+  message(FATAL_ERROR "Cannot find install manifest: @CMAKE_CURRENT_BINARY_DIR@/install_manifest.txt")
+endif(NOT EXISTS "@CMAKE_CURRENT_BINARY_DIR@/install_manifest.txt")
+
+file(READ "@CMAKE_CURRENT_BINARY_DIR@/install_manifest.txt" files)
+string(REGEX REPLACE "\n" ";" files "${files}")
+foreach(file ${files})
+  message(STATUS "Uninstalling $ENV{DESTDIR}${file}")
+  if(IS_SYMLINK "$ENV{DESTDIR}${file}" OR EXISTS "$ENV{DESTDIR}${file}")
+    exec_program(
+      "@CMAKE_COMMAND@" ARGS "-E remove \"$ENV{DESTDIR}${file}\""
+      OUTPUT_VARIABLE rm_out
+      RETURN_VALUE rm_retval
+      )
+    if(NOT "${rm_retval}" STREQUAL 0)
+      message(FATAL_ERROR "Problem when removing $ENV{DESTDIR}${file}")
+    endif(NOT "${rm_retval}" STREQUAL 0)
+  else(IS_SYMLINK "$ENV{DESTDIR}${file}" OR EXISTS "$ENV{DESTDIR}${file}")
+    message(STATUS "File $ENV{DESTDIR}${file} does not exist.")
+  endif(IS_SYMLINK "$ENV{DESTDIR}${file}" OR EXISTS "$ENV{DESTDIR}${file}")
+endforeach(file)
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -1,0 +1,136 @@
+project(gui)
+cmake_minimum_required(VERSION 3.0.2)
+FIND_PACKAGE(PkgConfig REQUIRED)
+include(GNUInstallDirs)
+
+set(GUI_HDR
+    Main.hpp
+    MenuBar.hpp
+    MainWidget.hpp
+    AddressBox.hpp
+    VideoDock.hpp
+    InfoDock.hpp
+    PlaylistDock.hpp
+    PlayClass.hpp
+    DemuxerThr.hpp
+    AVThread.hpp
+    VideoThr.hpp
+    AudioThr.hpp
+    SettingsWidget.hpp
+    OSDSettingsW.hpp
+    DeintSettingsW.hpp
+    OtherVFiltersW.hpp
+    PlaylistWidget.hpp
+    EntryProperties.hpp
+    AboutWidget.hpp
+    AddressDialog.hpp
+    VideoEqualizer.hpp
+    Appearance.hpp
+    VolWidget.hpp
+)
+
+set(GUI_SRC
+    Main.cpp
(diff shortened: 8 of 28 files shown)


zaps166:
Many thanks for a big commit!

This is also good time to test QMPlay2 on *BSD systems. I haven't tested it since ~5 years :)

Issues which I found:
- LINGUAS -> LANGUAGES.
- Missing PortAudio: `USE_PORTAUDIO` (it should be disabled by default for Linux, enabled for other OSes).
- `USE_FFMPEG` - you shouldn't add an option to disable it, because QMPlay2 will also use the FFmpeg dependency in other modules (Equalizer, Visualizations, VideoFilters, libqmplay2). Also playlists won't load without the FFmpeg module, because it depends on the file reader which is in this module.
- At least Chiptune can be splitted int two submodules: CHIPTUNE_GME, CHIPTUNE_SID.
- "Build with qt5" -> "Build with Qt5"; also ALSA with capitals, and "PulseAudio" not "pulseaudio" in the tooltip.
- "Build with tags editor" - hmm, maybe notice about TagLib library in the tool tip?
- Is it possible to enable Qt5 by default only when Qt5 >= 5.6.1 (now already released) is detected? E.g. Qt 5.5.x and older are extremely buggy :smile: See a7406b968709db778e0187a061cb8b77299051bf and https://github.com/qtproject/qtbase/commits/5.6?author=zaps166
- `make uninstall` doesn't work (I know that CMake doesn't have it by default...).
- DirectX, FileAssociation - I think that nobody will compile QMPlay2 for Windows, but I'll try later whether it works :smile: It is easier for me to use `qmake` and scripts for automated Windows build with installer.
- Missing MAN page.
- Can't run QMPlay2 without installation (can't find modules) :( But this probably must be fixed in QMPlay2 code, right?
- Also you can add `USE_FFMPEG_AVDEVICE` (enabled by default for Linux) and `USE_AVRESAMPLE` (disabled by default, because SWRESAMPLE is default) :)
- Missing `USE_AUDIOCD`.
- Please add it to ChangeLog (`added CMake build (author)` or something like this).

Is it possible/easy to add all CMakeList files (except the file in the root directory) into the `/CMake/*` directories?

Is it possible to use release build by default? Then `-DCMAKE_BUILD_TYPE=Debug` for debug build, but release by default.

> I would like to update the Arch's PKGBUILD and Gentoo's ebuild to use the new CMake.

I can't fix Gentoo ebuild (not my package) and of course I will change the PKGBUILD :)

On this occasion maybe I should also change the `modules` directory. It should be in `/usr/lib/qmplay2/modules`, not in `/usr/share/qmplay2/modules` :smile:

arthurzam:
Thanks for the reply. I already started working on those issue. Do you prefer me adding a new fix commit or overwriting the previous one (personally I prefer overwrite for much cleaner and shorter history tree/line).

Building on Windows is broken (fix isn't straightforward at all), and I have no idea about MacOS.

I will incorporate the changes you suggested.
I'm not so sure about TagLib issue. I guess someone maybe won't need it and it reduces an dependency. I manage to run without it. If you talk about the documentation string, than please suggest a better one.
Also I didn't understand about USE_FFMPEG_AVDEVICE and USE_AVRESAMPLE .

I believe that using qmake during development is much better.
The issue about running directly without installing, it needs to be fixed in code (/src/gui/Main.cpp:397). But I don't think we really need to support this, because working in Qt Creator with CMake isn't very productive.

Do you want me to change the search path (/usr/share/ -> /usr/lib/ for the modules) in the code too?

Also, I plan to add CPack in the near future. This will make it easier to create packages (it creates DEB, RPM, NSIS, source tar balls, etc.). I have no experience with CPack, but I think it will work great.

zaps166:
> Building on Windows is broken (fix isn't straightforward at all)

So please don't do it (just remove everything for Windows). Compiling anything for Windows on Windows is a pain (especially in MSVC ^^, but also via MSys/MinGW). It is better to cross-compile using own scripts and modified AUR packages and MinGW/GCC :)
I'll try CMake on OS X later (on VirtualBox).

> Also I didn't understand about USE_FFMPEG_AVDEVICE and USE_AVRESAMPLE .

QMPlay2 (libqmplay2) supports also libavresample (disabled by default): https://github.com/zaps166/QMPlay2/blob/master/src/qmplay2/qmplay2.pro#L39

QMPlay2 (FFmpeg module) uses libavdevice for e.g. v4l2 playback on Linux (I'm using it sometimes for old TV card): https://github.com/zaps166/QMPlay2/blob/master/src/modules/FFmpeg/FFmpeg.pro#L42 (enabled by default only for Linux)

> If you talk about the documentation string, than please suggest a better one.

Yes, about tool tip and var name, maybe `OPTION(USE_TAGLIB "Build with tags editor" ON)`, but I don't know about `tags editor` or `tag editor` which is correct english form :smile:

> Do you want me to change the search path (/usr/share/ -> /usr/lib/ for the modules) in the code too?

Hmm... Ok, so I'll change the C++ code before merge :) It is good time to change it, because libs should be in `lib` directory.

> Also, I plan to add CPack in the near future.

Great (whatever it is, I don't know CPack :smile:)! But just don't do it for Windows (NSIS), it doesn't make any sense. I'll improve the Windows installer somewhere in the future. Currently my installer works properly with auto-updates :)

> Do you prefer me adding a new fix commit or overwriting the previous one (personally I prefer overwrite for much cleaner and shorter history tree/line).

What about adding next commits and then - if it will be ready - squash it and merge as one commit?


Btw. how does it behave e.g. if taglib devel files doesn't exists? Is it disabled by default or CMake fails or compilation fails?

And what about libdir? - some systems uses `/usr/lib64` for 64-bit libraries or Ubuntu have it somewhere else `/usr/lib/a_dir_with_64bit_libs` - I don't remember :smile:

arthurzam:
OK, I will remove Windows completely.

About the lib/lib64 - I use CMake's variable - which every distribution fills the information when they build CMake binary itself. Therefore in this issue we are portable.

About the missing devel - it will behave exactly aas if you called pkg-config. If it doesn't finds the libs or headers, it fails to compile with saying for example "x11 not found".

I will add new commits for now, and when we decide to merge I will squash all of them into one commit as you said.

zaps166:
Ok, so then disabling taglib makes possible to configure & compile.

On some CMake projects is shows summary, e.g.:
```
PulseAudio - Yes
ALSA       - Yes
PortAudio  - No
OpenGL     - Yes (OpenGL or OpenGL|ES)
VDPAU      - Yes
VAAPI      - Yes
SWRESAMPLE - Yes
AVRESAMPLE - No
XVideo     - Yes
...
```
Can it be done automatically in CMake? Or maybe it leads to adding a lot of scripts line and it isn't worth?

arthurzam:
I think that this summary is part of what the distribution adds by itself. I know that Gentoo and Arch does it.
It isn't hard to do, but do you think we need it? (Your decision here)

zaps166:
I thought about this: https://github.com/Itseez/opencv/blob/master/CMakeLists.txt#L751 , so this is distribution independent. Just simple prints only about which `USE_*` is enabled. Summary will be fine if `USE_*` will be auto-detected, but if user must disable it manually for missing devels, then is is not needed (user knows what he/she enables/disables), so don't do it. Maybe another time :)

arthurzam:
I fixed most of the issues, as explained in the commit's message (in the commit squash the message will be cleaned - for now I use them to easier distinguish changes).

I cleaned many parts of Qt4/5 integration. Many of them were reused and now the AUTOMOC of Qt4 is working (I still prefer to check it in some Ubuntu distribution with Qt4).

About the Qt5 < 5.6.1 - Even though it might be buggy (I used quite some time with 5.4 with no problems), I decided to still use default Qt5 by show warning to the user that it is buggy. Mainly it will be packagers (or Gentoo users) that tweak those options, and this warning with do it job.

I changed the modules install path, and now it doesn't work if you install. But after adding symlink to the folder everything works. So this PR is ready for your change of path in code.

I added myself to the changelog, but I'm not sure about the date.
Also, I would like to ask you about the versioning of QMPlay2 - what is the major, minor number? If we use the date of release, then can I assume for example major.minor.patch is for example 2016.06.13 (year is major)?

zaps166:
Thanks!

- Missing word: "Build Chiptune with support" -> "Build Chiptune with SIDPLAY support".
- I think the icon should be in `share/icons/hicolor/128x128/apps`. Currently I'm creating a symlink for it in my script, but it can be only there, not directly in `share/icons`.
- For consistency in `lang/CMakeLists.txt` remove space after `if` and `endif` and `else` :smiley: also you can also remove duplicated end-lines in some files.
- MAN page is corrupted, it needs the version (see https://github.com/zaps166/QMPlay2/blob/master/compile_unix#L69)

> remove MacOS / Windows support code

Removed Windows support - very good, who cares about manual Windows compilation? :smiley: If someone wants to do it than it should do it manually or use cross-compiler and scripts :)

QMPlay2 has poor Mac OS X support, but it can run there, so anyway adding a few exceptions will be OK (I can test it, I've got OS X on VirtualBox). E.g. don't enable ALSA and PulseAudio, don't use MPRIS2 and DBUS, disable AudioCD by default.

> About make uninstall, I will need to add new file to a new folder named cmake, and no other file will be there (for now). Is it accepted?

Yes, but I think you can also add it into root directory (I think it is possible `First create file cmake_uninstall.cmake.in in the top-level directory of the project`) or into `src` directory :)

Btw. Is it possible to add QMPlay2 directory structure and `CMakeList` files into `/CMake` directory? Or it won't work properly or too many work?

It is also good time to test it under PC-BSD (simplified FreeBSD to not waste time for installing the real FreeBSD).

I'll change the code soon for `/usr/lib/qmplay2/modules` :)

CMake `3.0.2` or higher is necessary?

> I used quite some time with 5.4 with no problems

I depends what do you do with it. Example is when you disconnect/connect a screen (e.g. external screen on laptop or if your HDMI cable (doesn't have screws :/) in PC will plug-out then in Qt5 < 5.6.0 on X11 every application can crash or its GUI may disappear - main Qt5 BUG before 5.6.0 which I fixed...). And many missing mouse enter/leave events, especially on X11 (Qt5 < 5.6.1) which causes incorrect cursor shape in some places, mainly in text editor (like QtCreator) or when moving pages in Okular without releasing mouse button (which was extremely annoying for me and also for other users), etc... Btw. I found another small regression Qt 5.6.x -> Qt 5.7.0 and it will be fixed in Qt 5.7.1 :disappointed:

zaps166:
After your great work I think that I can remove my installation script (or change it to CMake commands) :D

arthurzam:
I updated the strings.

About the man page - I decided that when we use the man page, we change a variable in there  (the version), therefore to distinguish I added .in extension - making it obvious we will change in deployment. I updated the variable to be "flashy" and updated the compile_unix script to fix this change.

The icon is now installing into the proper location (without the one in icons folder).
What about those KDE solid actions - I don't know anything. And we can't check if one folder exists (hell know in what system it will be script built and in what sandbox) - we need to somehow choose one.

About CMake 3.0.2 - for much better Qt4/5 integration, much cleaner system, it is suggested to use CMake 3.0, when the used version is 3.0.2 (bug fixes). This is quite an old version (here using CMake 3.5.2) when the stable version is 3.4 - therefore it isn't bleeding edge.

About the whole deal with MacOS and BSD - I can add checks for them, but I'm not so familiar with the concepts there. For adding support we will need which packages work, which modules should be unsupported (like DirectX is only for Windows). Therefore I will need your full instructions here.
Sadly, I can't install any virtual machine on my machines cause my GPU died ...

Also, as a constant user of Qt5, I would like to thank you also for even patching/reporting bugs in this toolkit. The community needs more people like you!

zaps166:
> About the man page (...)

OK! (unnecessary backslash `\` before first `@` in script?)

> The icon is now installing into the proper location (without the one in icons folder).

OK

> What about those KDE solid actions (...)

I don't know. Currently I don't have KDE4 (Arch doesn't have it, I switched to fusion of Mate+Xfce4+KF5). Maybe drop this (this is only for AudioCD for KDE4)? I don't know where it should be placed (see script, it checks two directories) and how to detect KDE4 solid actions path in CMake. Probably this won't work in Plasma5 - not tested.

> About CMake 3.0.2 - for much better Qt4/5 integration (...)

OK

> About the whole deal with MacOS and BSD

I'll try this and next I'll write how does it work there. Currently I know this:
- On OS X: ALSA and PulseAudio and DBus(MPRIS2) should be unavailable. PortAudio enabled by default. On OS X support only for Qt5 (don't do any exception for OS X/Qt4). The `modules` directory like in my script (similar to Windows). 
- On *BSD: ALSA should be unavailable. PortAudio enabled by default. Everything else like on Linux (similar OS).

> I can't install any virtual machine on my machines cause my GPU died ...

I don't understand this sentence. VM causes GPU to die, or your GPU already died and you are using VNC? :)

> Also, as a constant user of Qt5, I would like to thank you also for even patching/reporting bugs in this toolkit. The community needs more people like you!

Thanks :D

---

Questions:
- What about my custom XML files in `src/gui/Unix` (see installation script and/or AUR package)?
- `make uninstall` works :) But it doesn't remove empty dirs - I assume that it is Ok, right?

arthurzam:
MIME xml files are now installing.

zaps166:
Ok, I've pushed a commit which changes the `modules` path.

I'm doing also:
```bash
sudo update-desktop-database /usr/share/applications
sudo update-mime-database /usr/share/mime
```
for MIME types and *.desktop file. Does other CMake apps do this if they've got custom MIME types? Or maybe this  commands should be executed only from package installation script?

Is it possible to set the `rpath` after installation to `../lib` or `../lib64` (autodetect)?

arthurzam:
I rebased my fork with the new commit. Also I squashed for now the commit into one.

About the update-*-database: from what I see, the packager should include it in the post install stage.
In the install stage (make install), we should think as we are running in a sandbox with access only to some DESTDIR and we should copy all our files into this folder. Afterwards the DESTDIR is archived/copied by the package manager to the system. And then the packager calls (that way or another) those commands.

What you mean about the rpath?
If you mean ELF's rpath, than installing the libqmplay2 (the only library linked, all the other are exe or modules) into the standard system's library dir (as you say lib lib64 lib32 etc) is all we need.
If you mean the location of the modules (when you load them in code), then we can add something like ifndef;define;endif block in which we ask from the build system the location or use the default one.

zaps166:
I meant the ELF executable `bin/QMPlay2`. Installing it to custom directory like `/home/user/something` needs setting the `LD_LIBRARY_PATH` manually (I thought that CMake can set the `rpath`, maybe additional option is possible?).

`Building QMPlay2 with Qt 5.5.x and older is extremely buggy` - this is OK, but sounds weird (needs adjustment) :smile: I assume that your english is much better than mine, so you can write a nice message (also 5.6.0 has too many problems with enter/leave events (worse than 5.5.x), maybe it mostly affects text editors and more advanced software, but it is still annoying. Also performance issues on 5.6.0 with FFT spectrum when window is large (not so important)) :) Maybe hmm... `Recommended Qt5 version is at least 5.6.1` or `Qt5 >= 5.6.1 is recommended for stable usage`?

Btw. e.g. w/o taglib I've got this: `Package 'taglib', required by 'world', not found` - what the `world` is :smiley: ?

I've started testing on PS-BSD...

zaps166:
Currently CMake finished w/o problems (can't execute `version` script due to `bash` in `/usr/local/bin`), but I've got: `/usr/bin/ld: cannot find -lswscale`

In CMake `link.txt` I can't find `-L/usr/local/lib` in linker flags. On FreeBSD most libraries are in `/usr/local/lib`. Maybe something is wrong with getting the PKGCONFIG library dirs?

It exists on my PC-BSD installation:
```bash
$ pkg-config --libs libswscale
-L/usr/local/lib  -lswscale
```

zaps166:
The same link problem with other libraries for modules. Maybe something is wrong with `${LIBSWSCALE_LIBRARIES}` and others?

arthurzam:
The problem was mine - missing link_directories (all those -L options). Now fixing.
About the version program - the problem is bash path or the script content itself?

Btw. can you please append at the end of the root CMakeLists.txt this snippet of code when you test on non Linux machines? I will need this results to distinguish the systems.

```
message("system = ${CMAKE_SYSTEM}")
message("system_name = ${CMAKE_SYSTEM_NAME}")
```

zaps166:
```
system = FreeBSD-10.3-RELEASE
system_name = FreeBSD
```

Please do a separate commit, otherwise I will not able to see the changes :)

arthurzam:
I'm not so sure, but from rumours I heard during development is that those "*BSD" systems have much more differences than the Linux distributions. There it is quite possible one combination in CMake for FreeBSD will be different than OpenBSD.
Therefore, I think it will be better if we add configuration options set per system, without using the wildcard *BSD.

zaps166:
Hmm, now is better (it compiles/links more things), but `/usr/bin/ld: cannot find -lswscale` e.g. in `Extensions` module. I don't get it why it needs `swscale`for linking this module. Maybe it gets the incomplete link flags form `libqmplay2`?

6de89bdbad7f11cf0560ad0eab0d63c0f40897c9: Not necessary, it is bug in my script, I'll fix this :)

zaps166:
Modules links to `libqmplay2`, but they shouldn't link to libraries used by the `libqmplay2` if are not used in module.

arthurzam:
This patch is still better, it will find bash even if it isn't in the eight place.

About this link with libqmplay2 - I will check it. I didn't intend it to auto link dependencies.

zaps166:
The `version` script doesn't need `bash`, it should use simple `sh` interpreter which is in `/usr/bin` on *BSD (probably on all BSDs). Other scripts must use `#!/usr/bin/env bash`, so this CMake modification will now require `bash` which is not really needed.

> About this link with libqmplay2 - I will check it. I didn't intend it to auto link dependencies.

Ok!

zaps166:
Please also change in compilation script: `cp src/gui/Unix/QMPlay2.1.in app/share/man/man1` to `cp src/gui/Unix/QMPlay2.1.in app/share/man/man1/QMPlay2.1`

arthurzam:
I rebased again - so look at the last commit.
I found the problems with the depends, and fixed it (I think).

zaps166:
Compiles & links on FreeBSD :)

MAN in FreeBSD is in `/usr/share/man` and in `/usr/local/man` (without `share`, weird OS...). Probably CMake has a tool for testing this (I don't know).

CMake also uses ALSA on FreeBSD (it redirects to OSS and works, maybe compatibility layer with Linux). I think that PortAudio should be disabled by default only for Linux, and ALSA should be enabled by default only for Linux.

arthurzam:
I added a skeleton for supporting default options set per platform.

In the long run, I intend to create a better function and arguments to make it much readable and cleaner.

zaps166:
More cross-platform (ALSA is only for Linux, so enable it only for Linux. It can also works on FreeBSD as an OSS wrapper, but don't enable it on other platforms than Linux. If ALSA is disabled, enable PortAudio):
```cmake
if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
    set(DEFAULT_ALSA ON)
    set(DEFAULT_PORTAUDIO OFF)
else()
    set(DEFAULT_ALSA OFF)
    set(DEFAULT_PORTAUDIO ON)
endif()
```
and we can merge this :)

zaps166:
Please also add a commit message as a comment (I want to squash it and prevent unnecessary merge commit). Any other bug can be fixed anytime.

arthurzam:
I squashed all of them and changed the defaults as you said.

If you want to use the CMake in compile_unix, don't forget to append '-DCMAKE_INSTALL_PREFIX:PATH=/usr' to the cmake command (so it won't use /usr/local).

zaps166:
Thanks!
Do you want a merge commit? (I don't like merge commits for one commit, your decision :))

arthurzam:
I prefer without merge, cause it is rebased and ready and clean.

zaps166:
Ok, so I'll take a second option from Github.

zaps166:
One question: `OPTION(USE_QT5 "Build with Qt5" ON)` - what if Qt5 will be missing? Does it fallback to Qt4 automatically?

...and I hope that openSUSE won't use Qt5 in Packman repository for Qt5 <= 5.6.0 (it still uses Qt4) :)

arthurzam:
About Qt5 - No it won't. But you can easily pass an argument on the first call to CMake (you can also in QtCreator) and you will use Qt4 - it is an options (like passing on CONFIG to qmake).

About openSUSE, Ubuntu, etc. -> you or them will chose what options to enable and pass them during build script. Thats the whole point of CMake option.

zaps166:
So OK then, merging. Everything can be fixed later (still not tested on OS X, MAN is in wrong dir on FreeBSD, etc.) :)

Issue #22, opened 2016-06-16
CMake improvements


zaps166:
Task for PR #32 (DONE):
- [x] Support for static linking for cross-compilation for Windows

Task for PR #31 (DONE):
- [x] CMake support for cross-compilation for Windows

Task for commit 2b9e1aa3c8c710fe4c8e169693592769588c9fd7 (DONE):
- [x] CMake support for OS X

Task for commit 4e30dc47e14bbae15b3d7b12dfa64c61c356feb7 (DONE):
- [x] detect OpenGL|ES for Qt4 and set compiler macros for it

Task for PR #28 (DONE):
- [x] show Solid actions in summary
- [x] review scripts and do small cleanup
- [x] use `pkg-config` for X11 in XVideo
- [x] detect OpenGL|ES 1.0, stop CMake and show error
- [x] option to keep RPATH after install (for `QMPlay2` executable to `libqmplay2.so`)

Task list for PR #24 (DONE):
- some modules should be optional (enabled by default only when the dependency exists):
 - [x] GME (libgme; on \*Ubuntu `pkg-config` doesn't exists for this, so detect it like in `*.pro` file)
 - [x] SID (libsidplayfp)
 - [x] PulseAudio (libpulse-simple)
 - [x] AudioCD (libcdio and libcddb)
 - [x] XVideo (libX11 and libxv)
 - [x] VA-API (libX11 and libva and libva-x11)
 - [x] VDPAU (libX11 libvdpau)
- [x] show warning when above modules will be disabled due to missing dependencies
- [x] show summary which modules will be built (needed for the previous point)
- [x] fallback to Qt4 if Qt5 doesn't exists
- [x] fallback to Qt4 if Qt5 is older than 5.6.1 instead of warning (if packager wants to use Qt5 < 5.6.1, then he/she can just enable it from command line `-DUSE_QT5=Yes`)
- [x] do something with MAN page for FreeBSD (it is installed to incorrect directory in `/usr/local`)
- [x] install the license
- [x] install ChangeLog, TODO, README.md
- [x] `make uninstall` should remove empty directories
- [x] detect the FFmpeg version (FFmpeg 2.2.0 or newer is needed)
- [x] install `QMPlay2-opencda.desktop` into proper directory (see e8d60d936789df0abbd45ebecfadf3182af15b35)
- [x] allow to disable ProstoPleer via CMake (enabled by default, see 9aac5d112681094a3e1429896bc122d0eff73891)
[...]

PR #28, opened 2016-06-20
CMake: Small cleanups, minor fixes

zaps166:
Major changes:
- use pkg-config in XVideo module,
- show Solid actions in summary,
- option to set install RPATH,
- error on OpenGL|ES 1.0.

Task #22


zaps166:
@arthurzam Hi! could you look at this PR? Thanks!
Is it possible to set a variable and show changes in `cmake-gui` (e.g. the auto-detected Solid actions path)?

arthurzam:
> /src/gui/CMakeLists.txt

The code for UI files should still be there. I saw that you would like to move settings forms from code to ui file (do you need help?), and maybe in the future you would add some new forms. This makes it ready to be added at any moment.

Everything else looks OK for me :)

zaps166:
> The code for UI files should still be there.

I can add it again, but Ok, I'll revert it (I want to move some GUI parts to `*.ui`)

Also removed policy from `/CMakeLists.txt` - doesn't seem to be necessary in this file.

> do you need help?

I'll think about moving GUI later, so I'll notify you :)

> About the task "CMake support for cross-compilation for Windows" - with what toolchain do you want to use it?

Is it tool-chain dependent? If yes - `{i686,x86_64}-w64-mingw32` from Arch Linux.

PR #31, opened 2016-06-23
Cross Compile CMake support

arthurzam:
As with todo list on #22 

I was compiling with mingw-w64 of Arch Linux. I needed to install those applications (latest version):

1. mingw-w64-cmake
2. mingw-w64-ffmpeg
3. mingw-w64-gcc
4. mingw-w64-portaudio
5. mingw-w64-qt5-base

Therefore, I have no idea what happens on old versions of those packages.

Some changes I have done:

- added option to specify LANGUAGES as empty string - same result as with None.
- added CMake script file for DirectX and FileAssociations.
- In FileAssociations I needed to add the UNICODE definition for it compiling. something with converting QString to LCSTR (maybe we should make in code ifndef UNICODE?)
- some more checks for libraries (find zlib for Modplug)
- about OpenGL module: there were some gl functions that I needed to find the appropriate lib file (wasn't specified in pro file). After binary grep in all mingw folder the only one that worked was GLESv2. Please say your thoughts!
- I managed only to compile. I can't use Wine and I don't have Windows to check. So testing if it works is on you. Also can you share with me Window's files list?

On another note - I'm researching now how to install with program all of its dependencies. For example including with PortAudio module the portaudio dll. In case I manage it it will make packaging for Windows easier (I think).

PR #32, opened 2016-06-25
CMake cross compile static support

arthurzam:
As direct continuation of PR #31 and task list #22 .

Added support for static linking AudioCD, Chiptune, PortAudio, libqmplay2, gui. This consists of an advanced string option, named `STATIC_*_LIBRARIES `, which replaces the linked libraries. Checked using `-DSTATIC_PORTAUDIO_LIBRARIES="libportaudio.a libwinmm.dll.a" -DSTATIC_GUI_LIBRARIES="libtag.a libz.dll.a"`. After checking the dll dependencies (using `objdump -p <file> | grep DLL`) it is working.

Added Updater to Windows.

Default build type should be updated. About the option of solid-actions - I think is is better to not update in GUI cause empty value shows to auto detect.


zaps166:
Thanks!

> Added Updater to Windows.

Ok, I forgot about this yesterday :smiley:

> Default build type should be updated. About the option of solid-actions - I think is is better to not update in GUI cause empty value shows to auto detect.

It discards the default description. Maybe don't change this (don't write the description) or re-use the default description if possible.

> STATIC_*_LIBRARIES

This allows to specify also linking to DLL's, so maybe just name it `*_LIBRARIES`? In this case add also advanced boolean option `GUI_STATIC_TAGLIB` for definition.
Maybe do it for all OS'es, not only for Win32?

zaps166:
Also `STATIC_QMPLAY2_LIBRARIES` -> `LIBQMPLAY2_LIBRARIES` (add `LIB` prefix).

zaps166:
Works very good, but it shows the CMD by default (missing linker flag `-Wl,-subsystem=windows` - please add advanced only for Windows `USE_CMD`, disabled by default), my command line:
```sh
i686-w64-mingw32-cmake .. -DUSE_QT5=NO -DCMAKE_INSTALL_PREFIX=~/Desktop/QW32 -DSTATIC_AUDIOCD_LIBRARIES="libcdio.a libcddb.a libregex.a libwinmm.dll.a libws2_32.dll.a" -DSTATIC_GUI_LIBRARIES="libtag.a libz.dll.a" -DSTATIC_PORTAUDIO_LIBRARIES="libportaudio.a libwinmm.dll.a" -DSTATIC_QMPLAY2_LIBRARIES="libswscale.dll.a libswresample.dll.a libavutil.dll.a libass.a libfontconfig.a libexpat.a libfreetype.a libfribidi.a libwinmm.dll.a libshell32.dll.a libpowrprof.dll.a" -DSTATIC_CHIPTUNE_LIBRARIES="libgme.a libsidplayfp.a" -DCMAKE_CXX_FLAGS_RELEASE="-O2 -DNDEBUG -fno-exceptions -g0" -DCMAKE_C_FLAGS_RELEASE="-O2 -DNDEBUG -g0"
```
```sh
make install/strip
```
:smiley:

I think that we should disable exceptions to reduce size (`-fno-exceptions`) - QMPlay2 doesn't use this. Also maybe use `-O2` by default for release build? Also why it adds `-g` for windows (maybe default CMake config from Arch Linux package?) . Maybe force `-g0` for release build as default?

arthurzam:
> rename options to *_LIBRARIES

I'm against it because it might make problems with pkg-config finders. The pkg-config cmake module is quite advanced. It it sees this variable it will stop thinking that the results are already in cache, and that way we won't get the definitions & include libraries. Therefore I think this is better.
> advanced only for Windows USE_CMD

will add in next commit
> disable exceptions

I will add if you say so. I only want to inform you on something. Many users of QMPlay2 (like me) use the git version. As programming is unstable (maybe you will break something somewhere in code), we need to understand what happened. Without exceptions I think there would be just some global fault and a crash. Therefore I think disabling exceptions on Release (like the last one is '16.06.01') is better. Or I can add an advanced option.
> default -O2 and -g for Windows

CMake by default uses -O3 for release build. The wrapper adds the -g flag. In yours build script I suggest to not use this wrapper, and copy the cmake invocation lines, and manage them. This wrapper is quite useful, but not the best for release compilation.

zaps166:
> I'm against it

So `CUSTOM_*_LIBRARIES` :)

zaps166:
> Without exceptions I think there would be just some global fault and a crash.

Hmm, I think that if you want a backtrace you shouldn't strip binaries. `-fno-exception` doesn't allow you to use `try{}catch(){}` and `throw`, so it also reduces size (QMPlay2 never uses it, but Qt can use it sometimes if enabled). Maybe just read QT_CONFIG or something and use the flag which was used in Qt compilation?

zaps166:
> CMake by default uses -O3 for release build. The wrapper adds the -g flag. In yours build script I suggest to not use this wrapper, and copy the cmake invocation lines, and manage them. This wrapper is quite useful, but not the best for release compilation.

So I'll use `-DCMAKE_C_FLAGS_RELEASE` and `-DCMAKE_CXX_FLAGS_RELEASE` :)

arthurzam:
> about USE_CMD

After searching I found that the eight version for GCC is [-mwindows](https://gcc.gnu.org/onlinedocs/gcc-4.6.2/gcc/i386-and-x86_002d64-Windows-Options.html). Ready in the commit.
> rename to CUSTOM_*_LIBRARIES

done.
> exceptions

So, I will ad this option if clang or gcc.

zaps166:
> So, I will ad this option if clang or gcc.

Or maybe is it possible to read the flag from qt configuration?

arthurzam:
I have no idea how to read qt configuration from cmake. Another option is to add it to release flags.

zaps166:
> I have no idea how to read qt configuration from cmake. Another option is to add it to release flags.

${QT_CONFIG}, works with Qt4, but not tested for Qt5 (see OpenGL2 module)

zaps166:
Or maybe it should be in Qt5 here `${Qt5Widgets_EXECUTABLE_COMPILE_FLAGS}` or in similar variable ?

arthurzam:
Why not make it user's responsibility ?

zaps166:
> Why not make it user's responsibility ?

User always can use `-DCMAKE_CXX_FLAGS` to override. `qmake` uses the Qt configuration flags by default (also for exceptions).

zaps166:
Maybe it already works for Qt5 (?) and it needs only set this in Qt4 (my Qt4 build doesn't use exceptions)?

arthurzam:
I know that the wrapper added exceptions. Check on regular cmake invocation.

zaps166:
`Static libs for GUI` -> `Custom libraries for GUI` - change for all modules :)

zaps166:
`add_definitions(-DTAGLIB_STATIC)` please make it configurable by `TAGLIB_STATIC_LINK`, disabled by default (CUSTOM_* doesn't mean that this will be static) :)

arthurzam:
changed the docstings for static libs
added -fno-exceptions for all release build (for now until we find for Qt5).
added TAGLIB_STATIC_LINK as you said.

zaps166:
Thanks!
I'll push some commits before merge into this PR later, OK?

zaps166:
I've reverted the "Release" build type (no need for it, it's better to have original description). I'll merge it soon :)

zaps166:
Thanks for the PR!

Issue #13, opened 2016-05-30
Move settings UI into QtDesigner


zaps166:
Modifying the settings window is very inconvenient in pure C++, so move settings UI into `*.ui` files.

arthurzam:
I would like to work on this task. But for this I have some questions:

1. In my opinion, for easier managing of the various object with names, it is better to create a ui file per every tab (or page as it is called in the code). To every ui file an appropriate class file will be created for managing the information associated with it (making it much more grouped).
2. `*.ui` files don't handle something like [ifdef](https://github.com/zaps166/QMPlay2/blob/master/src/gui/SettingsWidget.cpp#L477). I can fix this problem by setting the hidden flag when the ifdef happens.
3. ui files don't support partial translatability. For example you have lines like `(tr("Style") + ": ")`. Should I make it just as "Style: " or set the text in C++ code?
4. For grouping of files, should we move all future files of Settings From to specific folder named "settings"?

zaps166:
I'll think about this later, but:

- I want to do one folder with all `*.ui` files,
- hidden flag is not the same - it can create visible small space in grid layout.

One form with all tabs or tabs in different UIs and merge it in C++ (C++ main settings window) - I don't know which will be better.

zaps166:
Done: 2dc9b610b70c4b347b2ab48884aba36053f17023 (#38, #39, #40), 961e533c9f17eea20b68cdb1dbf1bf93fb67da1e, dc3d73fcac766f93f0e75c31bf9a926d5fb60ad4

@arthurzam Thanks!!!

arthurzam:
@zaps166 Was happy to contribute!

PR #35, opened 2016-07-01
Settings UI files

arthurzam:
About the #13 issue, I made some decisions:

1. I translated 'Strona' into 'Page' (checked using Google Translate :) )
2. Only create ui file for page 1 & page 2. Those two pages were enormous, therefore it is logical to create ui file for them. But the other pages are little / intuitive therefore I saw no need to create for them a ui file (There is a little overhead during initialization in using the generated code than yours).
3. I changed a little the concept in arranging the objects. Instead of QGridLayout I used combination of QFormLayout, QHBoxLayout and QVBoxLayout. In my opinion it is much cleaner looking now. But please says your thoughts.

Some more questions:

1. I think that moving every page (or at least page 1 & 2) into separate class and connect the signal/slot there. There will be only little more size addition to the binary, but it will make better organization.
2. The [OSDSettingsW](https://github.com/zaps166/QMPlay2/blob/master/src/gui/OSDSettingsW.hpp) is also quite long - should I also convert it into ui file?

**Reminder**: If you try to use/push to my fork, please do it into the `ui` branch. In the master branch there is future code (not ready still) for CPack, NSIS & OS X bundle.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -66,6 +66,8 @@ set(GUI_SRC
 
 set(GUI_FORMS # *.ui files
     ui/settings_general.ui
+    ui/settings_playback.ui
+    ui/settings_playback_moduleslist.ui
 )
 
 set(GUI_RESOURCES
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -65,7 +65,7 @@ set(GUI_SRC
 )
 
 set(GUI_FORMS # *.ui files
-
+    ui/settings_general.ui
 )
 
 set(GUI_RESOURCES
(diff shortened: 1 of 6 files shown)


zaps166:
Thanks! I'll look into this soon :)

zaps166:
Please retarget to `settings-ui` branch. Also please change `ui`->`Ui` directory name and change `*.ui` names to CamelCase (e.g. `SettingsGeneral.ui`, etc).

Coding style: CamelCase (classes, file names), camelCase (variable and function/method names) and stars `*` to the right. `page1` can be still `page1` instead of `page1_cont`. Also spaces after `for ()`, `if ()`, `while ()`.

Example for https://github.com/arthurzam/QMPlay2/blob/80dd82efbab68b0d2b0d11523b5c635d5ef9a65a/src/gui/SettingsWidget.cpp#L395 (don't mix spaces with tabs - it looks bad if editor has different tab spacing):
```c++
	const QString modulesTitleList[3] =
	{
		tr("Video output priority"),
		tr("Audio output priority"),
		tr("Decoders priority")
	};
```

> I think that moving every page (or at least page 1 & 2) into separate class and connect the signal/slot there. There will be only little more size addition to the binary, but it will make better organization.

Let's do it in one class (i.e. don't change) :)

> The OSDSettingsW is also quite long - should I also convert it into ui file?

Hmm, I don't know :smile: First I want to merge it and do some changes :) Maybe do it for all settings forms (except modules)?

> I changed a little the concept in arranging the objects.

Ok, but I think I should do more changes :) The spacing is too big, but don't change, I'll do it.

arthurzam:
Well, thanks for the reply.
Because of the retarget, I will need to close & reopen the pull request to a new one.

PR #38, opened 2016-07-03
Settings Ui

arthurzam:
Retarget of #35.

Fixed all the comments.

Also, about the styling problem. I don't think this problem won't happen again in future contributions. Maybe we should add some kind of style fixer script (for example using `astyle` or `clang-format`).

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -65,7 +65,9 @@ set(GUI_SRC
 )
 
 set(GUI_FORMS # *.ui files
-
+    Ui/SettingsGeneral.ui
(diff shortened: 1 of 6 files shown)


zaps166:
> Also, about the styling problem. I don't think this problem won't happen again in future contributions. Maybe we should add some kind of style fixer script (for example using astyle or clang-format).

I don't know these scripts :smiley:

zaps166:
Thanks!

All Settings Ui changes should be done in `settings-ui` branch. If everything will be finished be us - then merge to `master` & delete `settings-ui` branch.

arthurzam:
Understood.

PR #39, opened 2016-07-06
Settings Ui OSDSettings

arthurzam:
The next PR about issue #13 :)

This PR includes `OSDSettings` ui file (it was quite big and now it is a little better). As Page 4 & 5 use this widget, I created a little "hack" that enables those pages to add widgets after the widget and still look good (direct access to the VBox).

After some inspection, I think that Page 3 & 5 & 6 should not be converted to ui file (the UI code is too simple). Page 4 is around creating a simple GroupBox and appending it to OSDSettings, therefore I don't think it is worth moving to ui file.

I do see a difference in margin, spacing between yours QGridLayout and mine approach. In my opinion (which might be wrong for most people) my approach is better because this extra spacing makes reading the settings a lot easier, and the auto resize after window resize is a lot cleaner.

The [last commit](https://github.com/zaps166/QMPlay2/commit/2dca495fc0b4ac4262d031b582839c7d68a00a7e) is a small change to remove pointers that are used only during initialization. I think it would be a good idea to do it, saving a little space. It would be much better to also remove more unused pointers, but there would be a need to change the names, like changing `page6->layout` to `page6_layout`. But for this change I prefer to hear your decision.

The last point: conflict between the branches. There is currently a conflict between master and settings-ui. I checked how to fix this conflict (by doing `git checkout settings-ui` -> `git rebase master`) is to remove all the master changes and remain the settings-ui changes. Also I think that we should do a periodical rebase to master. This rebase will need a `git push -f` but this branch is an internal dev branch therefore it won't be a problem if history changes for this branch.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -68,6 +68,7 @@ set(GUI_FORMS # *.ui files
     Ui/SettingsGeneral.ui
     Ui/SettingsPlayback.ui
     Ui/SettingsPlaybackModuleslist.ui
+    Ui/OSDSettings.ui
 )
 
 set(GUI_RESOURCES
--- a/src/gui/OSDSettingsW.cpp
+++ b/src/gui/OSDSettingsW.cpp
@@ -20,18 +20,9 @@
 
 #include <LibASS.hpp>
 #include <Main.hpp>
-
-#include <ColorButton.hpp>
 #include <Settings.hpp>
 
-#include <QDoubleSpinBox>
-#include <QFontComboBox>
 #include <QRadioButton>
-#include <QPushButton>
-#include <QGridLayout>
-#include <QGroupBox>
-#include <QSpinBox>
-#include <QLabel>
 
 void OSDSettingsW::init(const QString &prefix, int a, int b, int c, int d, int e, int f, double g, double h, const QColor &i, const QColor &j, const QColor &k)
 {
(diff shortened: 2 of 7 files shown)


zaps166:
Thanks!
I'll look into this later, together with #38

zaps166:
Rebased!

Please check which things in settings are missing after rebase (if any) and please fix up/down buttons in `SettingsPlaybackModuleslist.ui` - it should move up/down the submodule, currently it does nothing.

All other things I'll do, thanks :)

Can I later squash your two PR into one commit?

arthurzam:
Thank you very much.

About the problem you mentioned - sorry about this. The fix patch is [here](https://github.com/arthurzam/QMPlay2/commit/bb7d8bd4d63cb2292d6406b073238fc72e3b595d). I decided to change a little the approach, by adding information to every button as the int to the appropriate list. I hope this is accepted solution (I testing building on Qt4 & Qt5 and running on Qt5 - works).

About the squash. I have no problems. I prefer that the author of the squash commit would be me (as in git history I mean, easy checking using `gitk --all`), but it is just a preference. If you don't manage it, then just squash and commit.

zaps166:
> The fix patch is here.

You can do a PR with this :)

> About the squash. I have no problems.

Then I'll squash your 3 PRs as a one (you as an author) when I'll start working at this.

> I decided to change a little the approach, by adding information to every button as the int to the appropriate list.

The code with QVariant? I'll look at this if I'll start doing with this :)

PR #42, opened 2016-07-08
Add Contributors to About Window

arthurzam:
Add a tab for Contributors in About QMPlay2, which shows the AUTHORS file.


zaps166:
Thanks!

PR #44, opened 2016-07-10
How to add sharpness control to player-Windows 10 version

elegey:
I enjoy the player quality, and I would like to add sharpness control to video module. And the instructions how to do that.  Thanks for your help, Great Player.


zaps166:
Please open an "Issue" instead of "Pull request" which merges two branches :)

PR #46, opened 2016-07-12
Various Small Fixes

arthurzam:
This is a collection of small 3 fixes. Some more info about them:

1. patch c53d6dfcbc4689cc3001908d843f99b1cb31d228 - use like used in [here](https://github.com/zaps166/QMPlay2/blob/master/src/gui/SettingsWidget.cpp#L441). I think this will work better.
2. patch e5e2b29ab328473c8661ee220dfd43a8c0dd3faa - POSIX grep command has the `-q` switch which won't output. I think it is better than the redirection.
3. patch a35d0be2755d7d2ea20faeb93ce993b854e0b2b8 - use setRange instead of setMinimum with setMaximum (less lines, more readable, more optimized call).

Btw, if I will find some other small fixes like in this PR, do you prefer to send them as small patch file which you can commit or as a new PR?

### DIFF
--- a/src/gui/SettingsWidget.cpp
+++ b/src/gui/SettingsWidget.cpp
@@ -852,7 +852,7 @@ void SettingsWidget::applyProxy()
 		if (QMPSettings.getBool("Proxy/Login"))
 		{
 			proxy.setUser(QMPSettings.getString("Proxy/User"));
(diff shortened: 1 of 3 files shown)


zaps166:
Ok, thx :smile:

PR #47, opened 2016-07-14
Move to QFormLayout

arthurzam:
Many layouts in this project uses a QGridLayout with the format of two columns and while in the first one is a label and another widget on the second column. For those types of types of layouts Qt's documentation recommends using QFormLayout. It has a special function called `addRow` which enables to do the listed before easier. The best way to see the change is in commit 644ea6a316d4e9d1718ed775a67b5687a87469d2.

I changes various layouts (maybe there are more). Before deciding to change a layout I checked if the resulting binary size reduces (no need to manually create a QLabel for example), and if there is performance boost (as QFormLayout is a little simpler for Qt).
The size reduction varied, but the total reduction was around 4KB (on my machine). The performance improvement was around 2%-5% when loading the layout and resizing the containing window.

Also the UI changes were little to none. The modules once were a little easier to read. Appearance, DeintSettingsW and TagEditor didn't change at all for me.

As a result we have smaller size while better time performance, while the code is a little easier to read (less lines, no need to count rows, more grouping of information). Therefore I think this is a good PR to merge (after squashing the commits and removing the "reduces ..." description). 

Also I thought that it isn't about moving to `*.ui` files, therefore I think to target the master branch and not settings-ui.

Also I ask **not to merge** at least until tomorrow (Friday, 2016-04-15) because I want to check other layouts. But I think it is a good idea to start reviewing all those commits now.

### DIFF
--- a/src/modules/FFmpeg/FFmpeg.cpp
+++ b/src/modules/FFmpeg/FFmpeg.cpp
@@ -162,6 +162,7 @@ QMPLAY2_EXPORT_PLUGIN(FFmpeg)
 #include <Slider.hpp>
 
 #include <QGridLayout>
+#include <QFormLayout>
 #include <QGroupBox>
 #include <QCheckBox>
 #include <QComboBox>
@@ -183,15 +184,11 @@ ModuleSettingsWidget::ModuleSettingsWidget(Module &module) :
 	decoderVDPAUB->setCheckable(true);
 	decoderVDPAUB->setChecked(sets().getBool("DecoderVDPAUEnabled"));
 
-	QLabel *vdpauDeintMethodL = new QLabel(tr("Improving deinterlacing quality") + ": ");
-
 	vdpauDeintMethodB = new QComboBox;
 	vdpauDeintMethodB->addItems(QStringList() << tr("None") << "Temporal" << "Temporal-spatial");
 	vdpauDeintMethodB->setCurrentIndex(sets().getInt("VDPAUDeintMethod"));
 	if (vdpauDeintMethodB->currentIndex() < 0)
 		vdpauDeintMethodB->setCurrentIndex(1);
-
-	QLabel *vdpauHQScalingL = new QLabel(tr("Image scaling level") + ": ");
 
 	vdpauHQScalingB = new QComboBox;
 	for (int i = 0; i <= 9; ++i)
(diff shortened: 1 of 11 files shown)


src/modules/FFmpeg/FFmpeg.cpp
```diff
@@ -233,26 +227,21 @@ ModuleSettingsWidget::ModuleSettingsWidget(Module &module) :
...
+	vaapiLayout->addRow(allowVDPAUinVAAPIB);
+	vaapiLayout->addRow(tr("Improving deinterlacing quality") + ": ", vaapiDeintMethodB);
 
 	#ifndef HAVE_VPP
-		vaapiDeintMethodL->setEnabled(false);
 		vaapiDeintMethodB->setEnabled(false);
 	#endif
 #endif
 
-	/* Pospiesz się */
```
zaps166: Don't remove, translate to `/* Hurry up */` (a few lines below there is another `/**/` to close "section")

zaps166:
Thanks, nice PR! Also `master` is a correct branch :)

> Also I ask not to merge at least until tomorrow (Friday, 2016-04-14)

In Poland today is Thursday 2016.07.14 (do you have different calendar?) :)

arthurzam:
> In Poland today is Thursday 2016.07.14 (do you have different calendar?) :)

Sorry, my mistake. I wanted to write 2016-07-15. Sorry.

src/modules/FFmpeg/FFmpeg.cpp
```diff
@@ -233,26 +227,22 @@ ModuleSettingsWidget::ModuleSettingsWidget(Module &module) :
...
-	QGridLayout *vaapiLayout = new QGridLayout(decoderVAAPIEB);
-	vaapiLayout->addWidget(allowVDPAUinVAAPIB, 0, 0, 1, 2);
-	vaapiLayout->addWidget(vaapiDeintMethodL, 1, 0, 1, 1);
-	vaapiLayout->addWidget(vaapiDeintMethodB, 1, 1, 1, 1);
+	QFormLayout *vaapiLayout = new QFormLayout(decoderVAAPIEB);
+	vaapiLayout->addRow(allowVDPAUinVAAPIB);
+	vaapiLayout->addRow(tr("Improving deinterlacing quality") + ": ", vaapiDeintMethodB);
 
 	#ifndef HAVE_VPP
-		vaapiDeintMethodL->setEnabled(false);
```
zaps166: `vaapiLayout->labelForField(vaapiDeintMethodB)->setEnabled(false);`

zaps166:
Ok, so waiting for tomorrow :smile:

arthurzam:
I finished from my side. Now waiting for your reviews :smile:

zaps166:
I think it is OK :smile: If I'll find a bug, I'll fix it in `master`, thanks for the code simplification!

PR #49, opened 2016-07-18
More Code cleanup

arthurzam:
Well, I continue checking various parts of the code, trying to clean it, reorganizing and etc.

1. patch 211a58160a577614aee211352c77dec8e2eaf692 - print this new line as STATUS variable, making it better outputted when there are filters on CMake output (PRINT_ENABLED_FEATURES function uses STATUS internally).

2. patch 02ce6263841a167a45f9d7359d3fd92003377aba - the buttons, labels and layout are only needed as variables during initialization, therefore use them as local variable in ctor. Because after removing labels the `struct Controls` is only of `Slider*`, I changed the `controls` to array of `Slider*`. Thus it resulted in double pointer. This is "scary" only in the header file (I hope this is OK with you).

3. patch c845f818bc3a531d727f2cdb3da65265911ab013 - in the EntryProperties class, I move the layout widgets adding to be a part in the if/else block. Thus the amount of ifs that was there before was removed.

4. patch 456e8b002ddb926e4f7cd01df52da29320e201b4 - remove the `Q_OBJECT` from TextEdit. This macro adds a lot of Qt specific code for signals and slot in the same class support. Removing this macro doesn't do any regression from my tests. Only possible point is Qt4.

5. patch 7462d1e61f69b3ca28d5f54dd588bcaaa1dd6e5a - use QString::startsWith

The purpose of those various commits from the various pull requests is to optimize QMPlay2, make it's code cleaner, make the code shorter.

### DIFF
--- a/src/gui/EntryProperties.cpp
+++ b/src/gui/EntryProperties.cpp
@@ -56,6 +56,7 @@ EntryProperties::EntryProperties(QWidget *p, QTreeWidgetItem *_tWI, bool &sync, 
 	setWindowTitle(tr("Properties"));
 
 	QGridLayout layout(this);
+	int row = 0;
 
 	nameE = new QLineEdit;
 	nameE->setText(tWI->text(0));
@@ -72,9 +73,8 @@ EntryProperties::EntryProperties(QWidget *p, QTreeWidgetItem *_tWI, bool &sync, 
 
 		nameE->selectAll();
 
-		catalogCB = new QCheckBox;
+		catalogCB = new QCheckBox(tr("Synchronize with file or directory"));
 		catalogCB->setChecked(!pthE->text().isEmpty());
-		catalogCB->setText(tr("Synchronize with file or directory"));
 		connect(catalogCB, SIGNAL(stateChanged(int)), this, SLOT(setDirPthEEnabled(int)));
 
 		browseDirB = new QToolButton;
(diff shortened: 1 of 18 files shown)


src/gui/VideoAdjustment.hpp
```diff
@@ -45,13 +42,7 @@ private slots:
...
 	void reset();
 private:
-	QGridLayout *layout;
-	struct Controls
-	{
-		QLabel *titleL, *valueL;
-		Slider *slider;
-	} *controls;
-	QPushButton *resetB;
+	Slider **sliders;
```
zaps166: You can use one "*" here :)

src/gui/PlayClass.cpp
```diff
@@ -95,7 +95,7 @@
 #endif
 
 PlayClass::PlayClass() :
-	demuxThr(NULL), vThr(NULL), aThr(NULL),
+	demuxThr(NULL), vThr(NULL), aThr(NULL), aRatioName("auto"),
```
zaps166: Not necessary to moving only one variable here. Also moving everything here won't look fine (I think)...

src/gui/EntryProperties.cpp
```diff
@@ -56,6 +56,7 @@ EntryProperties::EntryProperties(QWidget *p, QTreeWidgetItem *_tWI, bool &sync,
 	setWindowTitle(tr("Properties"));
 
 	QGridLayout layout(this);
+	int row = 0;
 
 	nameE = new QLineEdit;
```
zaps166: `nameE = new QLineEdit(tWI->text(0));`

arthurzam:
Done. Thank you.

About the `Slider**` -> `Slider*` - just for easier readability I prefer `&sliders[i]` than `sliders + i` (even though it is exactly the same under the hood). 

I moved only the QString initializer in `PlayClass.cpp` because this is a complex class (integers, doubles, char, ptr are simple). Before the change, an empty constructor was called for QString and than the assignment operator was called. Now the appropriate constructor is called.

zaps166:
> About the Slider** -> Slider* - just for easier readability I prefer &sliders[i] than sliders + i (even though it is exactly the same under the hood). 

Yes, you can also do `Slider *slider = &sliders[i]` and use `slider->` in the loop :)

arthurzam:
Added another patch (improves checks is string starts/ends with string).

> Yes, you can also do Slider *slider = &sliders[i] and use slider-> in the loop :)

I know it is possible, but I think the way it is now it is more readable.

zaps166:
> I know it is possible, but I think the way it is now it is more readable.

Hmm, avoid array in many lines should look better :)

```diff
diff --git a/src/gui/VideoAdjustment.cpp b/src/gui/VideoAdjustment.cpp
index 315dd0e..a19a732 100644
--- a/src/gui/VideoAdjustment.cpp
+++ b/src/gui/VideoAdjustment.cpp
@@ -59,17 +59,18 @@ VideoAdjustment::VideoAdjustment() :
 
                QLabel *valueL = new QLabel("0");
 
-               sliders[i].setProperty("valueL", qVariantFromValue((void *)valueL));
-               sliders[i].setTickPosition(QSlider::TicksBelow);
-               sliders[i].setMinimumWidth(50);
-               sliders[i].setTickInterval(25);
-               sliders[i].setRange(-100, 100);
-               sliders[i].setWheelStep(1);
-               sliders[i].setValue(0);
-               connect(&sliders[i], SIGNAL(valueChanged(int)), this, SLOT(setValue(int)));
+               Slider *slider = &sliders[i];
+               slider->setProperty("valueL", qVariantFromValue((void *)valueL));
+               slider->setTickPosition(QSlider::TicksBelow);
+               slider->setMinimumWidth(50);
+               slider->setTickInterval(25);
+               slider->setRange(-100, 100);
+               slider->setWheelStep(1);
+               slider->setValue(0);
+               connect(slider, SIGNAL(valueChanged(int)), this, SLOT(setValue(int)));
 
                layout->addWidget(titleL, i, 0);
-               layout->addWidget(&sliders[i], i, 1);
+               layout->addWidget(slider, i, 1);
                layout->addWidget(valueL, i, 2);
        }
 
```

What do you think?

arthurzam:
Applied.

src/gui/Updater.cpp
```diff
@@ -165,7 +165,7 @@ void Updater::writeToFile()
 	bool err = false;
 	if (firstChunk)
 	{
-		err = (arr.left(2) != "MZ");
+                err = (!arr.startsWith("MZ"));
```
zaps166: Drop `()`

src/qmplay2/Functions.cpp
```diff
@@ -156,7 +156,7 @@ QString Functions::fileName(QString f, bool extension)
 	if (f == "file:///")
 		return "/";
 #endif
-	while (f.right(1) == "/")
+	while (f.endsWith("/"))
 		f.chop(1);
 	QString n = f.right(f.length() - f.lastIndexOf('/') - 1);
```
zaps166: `const` ?

zaps166: `const` ?

arthurzam:
> Drop ()

Sorry, I don't understand?

zaps166:
> Sorry, I don't understand?

```C++
err = !arr.startsWith("MZ");
```

also use tabulators instead of spaces :)

arthurzam:
Done

zaps166:
Ok, thanks! Looks correct :)

PR #50, opened 2016-07-20
Shortcuts Editor in Settings

arthurzam:
Well, after adding shortcuts changing using the config file, I thought to add UI for this.
I changed some approaches in shortcuts in MenuBar:

- added a new class `ShortcutHandler` (I hope this name is ok? )
- instead of setting default shortcuts in `MenuBar::init`, and then setting the shortcut for a Action, I moved all of this into `ShortcutHandler::appendAction` and there it is selected.
- I added simple `QTableView` with `ShortcutHandler` as it's model. This `QTableView` is the page7 widget.

I might have some problems with code style (as I added new code and not just replacing as for before). Please review it.

Also, I might have missed some actions, therefore I'm asking you to report the missing actions.

Some future insight:
I added `ShortcutHandler::revert()` and `ShortcutHandler::reset()` for future code of reverting the changes and returning to defaults. I didn't test it (therefore I added the TODO comment) but is should work properly.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -38,6 +38,7 @@ set(GUI_HDR
     VolWidget.hpp
     ScreenSaver.hpp
     ShortcutHandler.hpp
+    KeyBindingsDialog.hpp
 )
 
 set(GUI_SRC
@@ -66,6 +67,7 @@ set(GUI_SRC
     VolWidget.cpp
     ScreenSaver.cpp
     ShortcutHandler.cpp
+    KeyBindingsDialog.cpp
 )
 
 set(GUI_FORMS # *.ui files
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -37,6 +37,7 @@ set(GUI_HDR
     Appearance.hpp
     VolWidget.hpp
     ScreenSaver.hpp
+    ShortcutHandler.hpp
 )
 
 set(GUI_SRC
@@ -64,6 +65,7 @@ set(GUI_SRC
     Appearance.cpp
     VolWidget.cpp
     ScreenSaver.cpp
+    ShortcutHandler.cpp
 )
 
 set(GUI_FORMS # *.ui files
--- a/src/gui/KeyBindingsDialog.cpp
+++ b/src/gui/KeyBindingsDialog.cpp
@@ -1,0 +1,82 @@
+/*
+	QMPlay2 is a video and audio player.
+	Copyright (C) 2010-2016  Błażej Szczygieł
+
+	This program is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Lesser General Public License as published
+	by the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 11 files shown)


zaps166:
Thanks :) I'll modify this later. Please retarget to `kb` branch (it is better to ask for new working branch :smile:)

zaps166:
Or I'll take advantage of this code later. Currently I'm doing something else, so please wait (key bindings is a low priority for me) :)

arthurzam:
No problem :smile:
I knew that you might have something else you were working on, so I wasn't expecting you automatically move your attention to this feature. So I will wait until you can (I must inform that I won't be able to help for some weeks after 2016-08-02).

About asking for a new branch and various other things, can I just send you an email in the future? Of course bigger things will be managed using a Issue or a Pull Request.

Also, you have full access to my fork, and I create those changes in different branches in my fork, therefore if you will want to change something, you can always modify in my fork (just don't do `git commit --amend` after push).

zaps166:
> can I just send you an email in the future

Yes :smile: or Issue - don't need to close and create new PR for working branch :)

---

I think that it should be as additional window, not a Page 7 (just like "Appearance" in Page 1) :)

---

About C++:

use more `const` API whenever possible (constBegin(), const iterators, etc.) - maybe it doesn't improve performance in this case, because it is not a CPU-intensive loop (`const` avoids calling `detach()`), but looks better :)

don't use `Q_NULLPTR` - this is Qt5 only, so not defined in Qt4. Also Qt 5.7 uses only C++11, so `nullptr` is better if your software must use at least Qt 5.7. In QMPlay2 (Qt 4.7.0 and GNU++98 compatibility) - use just `NULL`. Currently I don't need C++11 in QMPlay2 (I don't use STL and Qt4 has no real benefits of this and I can live e.g. without lambdas), so it is better to not use it for compilation speed-up (Qt < 5.7.0) and compatibility. Maybe I'll use C++11 in QMPlay2 in the future.

spaces after `if (` and `switch (` and stars to the right, after a space (`<Type *>`, `Type *name`).

arthurzam:
Well, when you have time to work on this feature (just informing the status).

- Moved to a new Dialog.
- removed `Q_NULLPTR`
- using more const iterators
- the button is under "Set Appearance" in page 1
- fixed revert & restore
- the style should be fixed

zaps166:
Ok :smile:

`QLatin1Char` what for?

arthurzam:
> QLatin1Char what for?

I mainly copied code from my old project that was inspired by qpdfview. Removed.

zaps166:
OK!

zaps166:
Just tested, so:
- the button is too long (add empty label before it in form layout),
- 'apply' and 'restore default' shouldn't close the window (apply != ok).

:)

zaps166:
Or place the button over/under two clear buttons :smile:

zaps166:
Or I'll fix it later, so merge :smile:

zaps166:
Thanks!

arthurzam:
Well, thank you :smile: 
(I was just about to send a commit, no problem then).

zaps166:
Ok :smile:
I'll change something in this later (e.g. I want to use existing single-instance class QMPlay2GuiClass) :)

Currently it is merged and can be used by everyone :)

PR #52, opened 2016-07-29
LastFM: Allow to disable it by compiler macro

arthurzam:
Like with ProstoPleer use flag, but for LastFM.

Also fixed `#if USE_MPRIS2` into `#ifdef USE_MPRIS2`


zaps166:
`#if USE_MPRIS2` :smiley: Probably I forgot to test it - didn't compile due to missing macro value, right?
Thanks :)

PR #53, opened 2016-07-29
Typos fix & Settings convert usage

arthurzam:
1. patch b4ad182a34c7860ea993df220e0c73bcdee0d81d - Little code cleanup commit. Use the Settings `getBook`, `getInt`, ... instead of `get().toInt()`. Also, it might be a good idea to add an `toQColor()`, as it is used in various places and is quite logical to be used. Your thoughts?
2. patch bcb1e130d69494d5425ea8d0f8317f3437e39e17 - fix some typos in README.md (used spell check in vim).

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ QMPlay2 is a video and audio player. It can play all formats supported by FFmpeg
 Table of Contents
 =================
 
-* [Instalation](#instalation)
+* [Installation](#installation)
 * [YouTube](#youtube)
 * [Spherical view](#spherical-view)
 * [ALSA](#alsa)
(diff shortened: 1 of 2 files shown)


zaps166:
Thanks :) (btw. one regression, already fixed :smile:)

Hmm, `toColor()`, `toStringList()` - can be done, but this doesn't change anything (binary size, speed, etc) - only aesthetic change.

arthurzam:
Thanks for the fix (missed it - sorry).
About `toColor` and `toStringList` - yes it won't change the binary size, but it will improve the aesthetic of the code. I think code like `.value<QColor>()` isn't that good, and therefore if we can hide it behind a simple call of `getColor` it will make maintaining code and analysing it easier.

zaps166:
Ok, done: 3f69bb0d8935a0c25f227e598974cb7fd1a86403 :) Please review :smile:

arthurzam:
Looks ok.
Thanks! :smiley:

PR #54, opened 2016-07-30
CMake: use AUTORCC

arthurzam:
Use CMake's AUTORCC for better resources usage. Simplifies a little.

This resulted from the cross-compile bug discussed in #31 .


zaps166:
Thx!

PR #55, opened 2016-08-01
Code Cleanup

arthurzam:
This Pull Request is a collection of various code clean up:

1. patch bb502bedbc8c03ad98ead59f40d3742728fabd13 - better `QLabel` create and making YouTube::playOrEnqueue as a non slot function.
2. patch fccbad8c993c813d6d953d1ade5f7783ca6667b8 - set `QMPlay2Extensions::addressPrefixList` as a const function (quite logical from its purpose). Enables settings more const calls.
3. patch 8fc008001dd22e288bbc538e953846a5b691826c - adding const around many objects in the code.
4. patch 85292dca5974ce40b84d72c5fd6fb949efdeac23 - use `QString::arg` instead of string appending. Making the code a little bit more readable, smaller binary size and nearly none performance change (changed between 2% more or 2% less which is negligible).
5. patch 932427b22b61875c5a4682c0da01b0b44b1c2f71 - like the previous one. Also use more define string than `static QString`.
6. patch b220140e7ecb6ea1d309ef61aa75c344aea9c6f7 - like the previous one.
7. patch 1177fa054f6cd286a7a8603d080fe8d95f2486af - like the previous one.
8. patch b92441b7c8374f4c42a5f63466e9e822ad8b2e67 - some `if` change and more const.

Most of the code changes are trivial, but I'm asking you **to check** the `QString::arg` changes (4-7). Even though I checked the changes, I might have missed something (mistake here might be a critical one).

The difference between before and after this patch set is around 10KB on my machine (Gentoo + GCC 5.4).

---

On another note about the future: from tomorrow onwards I won't be able to help or answer (except sometimes during Saturday and maybe Friday) as I'm enlisted to the army for the following three years. So no problems to send requests or questions, but please know that it will take time for me to answer.

### DIFF
--- a/src/gui/AddressBox.cpp
+++ b/src/gui/AddressBox.cpp
@@ -43,7 +43,7 @@ AddressBox::AddressBox(Qt::Orientation o, QString url)
 				pB.addItem(icon, mod.name, MODULE);
 			}
 
-	foreach (QMPlay2Extensions *QMPlay2Ext, QMPlay2Extensions::QMPlay2ExtensionsList())
+	foreach (const QMPlay2Extensions *QMPlay2Ext, QMPlay2Extensions::QMPlay2ExtensionsList())
 		foreach (const QMPlay2Extensions::AddressPrefix &addressPrefix, QMPlay2Ext->addressPrefixList())
 			pB.addItem(QPixmap::fromImage(addressPrefix.img), addressPrefix, MODULE);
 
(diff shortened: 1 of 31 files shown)


zaps166:
Ok, thx!
I'll review it later :)

zaps166:
I've used most cleanup from this PR in my commit. I don't want defines (even if it is better - strings can be merged in compile time). Some `.arg()` changes in this PR takes `int` as a second argument - this won't work as expected. Also I don't want regressions (I don't want to test every change) :smile: Thanks for cleanup!

PR #61, opened 2016-10-28
Add Notifications Support

arthurzam:
This Pull Request isn't finished yet in my idea, but I request your review. Please note that my schedule is very problematic, therefore If I don't reply then I can't reply at that time, but when I will have the ability I will fix and reply :smile: 

The idea in this PR is to add notifications support to QMPlay2, for example to show the title of the playing video when it moves to the next one.

Supports X11 notifications (like with notify-send) and system tray notifications.
Supports some settings changes, like format, timeout and what notifications to show. If you have more ideas of events to show notification please inform me.

Some questions for improvements:
1. Are the class names and Module name OK?
2. Is the code style fine (I checked many times to match yours but maybe I missed some place)?
3. Windows supports need testing (I can't compile now to test over Windows). The support should be using the system tray.
4. Should I move some code from one file to a new file (for example look at `Notifys.cpp` & `Notifys.hpp`)?
5. Use flags: should I add use flags (in CMake) for `USE_NOTIFY` and `USE_DBUS` (about this one I know that many users don't want dbus at all and therefore remove X11 notifications)?

### DIFF
--- a/src/gui/MainWidget.cpp
+++ b/src/gui/MainWidget.cpp
@@ -154,6 +154,7 @@ MainWidget::MainWidget(QPair<QStringList, QStringList> &QMPArguments)
 	tray = new QSystemTrayIcon(this);
 	tray->setIcon(QMPlay2Core.getIconFromTheme("QMPlay2-panel", QMPlay2Core.getQMPlay2Pixmap()));
 	tray->setVisible(settings.getBool("TrayVisible", true));
+	QMPlay2Core.systemTray = tray;
 
 	setDockOptions(AllowNestedDocks | AnimatedDocks | AllowTabbedDocks);
 	setMouseTracking(true);
--- a/src/modules/Notifies/Notifies.cpp
+++ b/src/modules/Notifies/Notifies.cpp
@@ -25,13 +25,11 @@ Notifies::Notifies() :
 {
 //	moduleImg = QImage(":/???"); // TODO: add icon
 
-	init("TypeDisabled", false);
+	init("TypeDisabled", true);
 #ifdef Q_OS_LINUX
-	init("TypeNative", true);
-	init("TypeTray", false);
-#else
-	init("TypeTray", true);
+	init("TypeNative", false);
 #endif
+	init("TypeTray", false);
 
 	init("timeout", 5000);
 
@@ -102,7 +100,7 @@ ModuleSettingsWidget::ModuleSettingsWidget(Module &module) :
 	connect(m_disabledR, SIGNAL(toggled(bool)), generalG, SLOT(setDisabled(bool)));
 
 	m_timeoutSB = new QDoubleSpinBox;
-	m_timeoutSB->setDecimals(4);
+	m_timeoutSB->setDecimals(1);
 	m_timeoutSB->setRange(0.0, 3600.0);
 	m_timeoutSB->setSingleStep(0.1);
 	m_timeoutSB->setSpecialValueText(tr("Infinite"));
--- a/src/modules/Notifies/Notifies.cpp
+++ b/src/modules/Notifies/Notifies.cpp
@@ -44,10 +44,8 @@ Notifies::Notifies() :
 
 QList<Notifies::Info> Notifies::getModulesInfo(const bool) const
 {
-	return QList<Info>()
-			<< Info(NotifyExtensionName, QMPLAY2EXTENSION);
+	return QList<Info>() << Info(NotifyExtensionName, QMPLAY2EXTENSION);
 }
-
 void *Notifies::createInstance(const QString &name)
 {
 	if (name == NotifyExtensionName)
--- a/src/modules/Notifies/Notifies.cpp
+++ b/src/modules/Notifies/Notifies.cpp
@@ -1,0 +1,171 @@
+﻿/*
+	QMPlay2 is a video and audio player.
+	Copyright (C) 2010-2016  Błażej Szczygieł
+
+	This program is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Lesser General Public License as published
+	by the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Lesser General Public License for more details.
+
+	You should have received a copy of the GNU Lesser General Public License
+	along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <Notifies.hpp>
+#include <Settings.hpp>
+#include <NotifyExtension.hpp>
+
+Notifies::Notifies() :
+	Module("Notifies")
+{
+//	moduleImg = QImage(":/???"); // TODO: add icon
+
+	init("TypeDisabled", false);
+#ifdef Q_OS_LINUX
+	init("TypeNative", true);
+	init("TypeTray", false);
+#else
+	init("TypeTray", true);
+#endif
+
+	init("timeout", 5000);
+
+	init("ShowVolume", true);
+	init("ShowTitle", true);
+	init("ShowPlayState", true);
+
+	init("CustomMsg", false);
+	init("CustomSummary", QString());
+	init("CustomBody", QString());
+}
+
+QList<Notifies::Info> Notifies::getModulesInfo(const bool) const
+{
+	return QList<Info>()
+			<< Info(NotifyExtensionName, QMPLAY2EXTENSION);
+}
+
+void *Notifies::createInstance(const QString &name)
+{
+	if (name == NotifyExtensionName)
+		return static_cast<QMPlay2Extensions *>(new NotifyExtension(*this));
+	return NULL;
+}
+
+Notifies::SettingsWidget *Notifies::getSettingsWidget()
+{
+	return new ModuleSettingsWidget(*this);
+}
+
+QMPLAY2_EXPORT_PLUGIN(Notifies)
+
+/**/
+
+#include <QBoxLayout>
+#include <QCheckBox>
+#include <QDoubleSpinBox>
+#include <QFormLayout>
+#include <QGridLayout>
+#include <QGroupBox>
+#include <QLineEdit>
+#include <QRadioButton>
+
+ModuleSettingsWidget::ModuleSettingsWidget(Module &module) :
+	Module::SettingsWidget(module)
+{
+	QGroupBox *notifyTypeG = new QGroupBox(tr("Notification Type"));
+	QVBoxLayout *notifyTypeL = new QVBoxLayout(notifyTypeG);
+
+	m_disabledR = new QRadioButton(tr("Disabled"));
+	m_disabledR->setChecked(sets().getBool("TypeDisabled"));
+	notifyTypeL->addWidget(m_disabledR);
+
+#ifdef Q_OS_LINUX
+	m_nativeR = new QRadioButton(tr("Show a Native Desktop Notification"));
+	m_nativeR->setChecked(sets().getBool("TypeNative"));
+	notifyTypeL->addWidget(m_nativeR);
+#endif
+
+	m_trayR = new QRadioButton(tr("Show a Popup from the System Tray"));
+	m_trayR->setChecked(sets().getBool("TypeTray"));
+	notifyTypeL->addWidget(m_trayR);
+
+	/**/
+
+	QGroupBox *generalG = new QGroupBox(tr("General Settings"));
+	generalG->setDisabled(sets().getBool("TypeDisabled"));
+	connect(m_disabledR, SIGNAL(toggled(bool)), generalG, SLOT(setDisabled(bool)));
+
+	m_timeoutSB = new QDoubleSpinBox;
+	m_timeoutSB->setDecimals(4);
+	m_timeoutSB->setRange(0.0, 3600.0);
+	m_timeoutSB->setSingleStep(0.1);
+	m_timeoutSB->setSpecialValueText(tr("Infinite"));
+	m_timeoutSB->setSuffix(" " + tr("sec"));
+	m_timeoutSB->setValue(sets().getInt("timeout") / 1000.0);
+
+	m_notifyVolumeB = new QCheckBox(tr("Show notification when volume changes"));
+	m_notifyVolumeB->setChecked(sets().getBool("ShowVolume"));
+
+	m_notifyTitleB = new QCheckBox(tr("Show notification when media changes"));
+	m_notifyTitleB->setChecked(sets().getBool("ShowTitle"));
+
+	m_notifyPlayStateB = new QCheckBox(tr("Show notification when play state changes"));
+	m_notifyPlayStateB->setChecked(sets().getBool("ShowPlayState"));
+
+	QFormLayout *generalL = new QFormLayout(generalG);
+	generalL->addRow(tr("Notification Timeout") + ":", m_timeoutSB);
+	generalL->addRow(m_notifyVolumeB);
+	generalL->addRow(m_notifyTitleB);
+	generalL->addRow(m_notifyPlayStateB);
+
+	/**/
+
+	m_customMsgG = new QGroupBox(tr("Use a custom message for media change notifications"));
+	m_customMsgG->setCheckable(true);
+	m_customMsgG->setChecked(sets().getBool("CustomMsg"));
+	m_customMsgG->setDisabled(sets().getBool("TypeDisabled"));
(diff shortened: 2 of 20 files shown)


zaps166:
Thanks for new module :)

Typo: `Notifys`->`Notifies`

I'll check it soon :smile:

arthurzam:
- Fixed this typo.
- Rebased over new changes.

I think future improvements are adding more notifications services. Do you know about OS X?

src/gui/MainWidget.cpp
```diff
@@ -154,6 +154,7 @@ MainWidget::MainWidget(QPair<QStringList, QStringList> &QMPArguments)
 	tray = new QSystemTrayIcon(this);
 	tray->setIcon(QMPlay2Core.getIconFromTheme("QMPlay2-panel", QMPlay2Core.getQMPlay2Pixmap()));
 	tray->setVisible(settings.getBool("TrayVisible", true));
+	QApplication::instance()->setProperty("system_tray", QVariant::fromValue<void *>(tray));
```
zaps166: Is `QApplication::instance()` better than `qApp` macro?
I think that you can create another variable for tray in `QMPlay2Core` class (without including `QSystemTrayIcon`). API/ABI compatibility is now not important.

src/modules/Notifies/Notifies.cpp
```diff
@@ -0,0 +1,175 @@
...
+*/
+
+#include <Notifies.hpp>
+#include <Settings.hpp>
+#include <NotifyExtension.hpp>
+
+Notifies::Notifies() :
+	Module("Notifies")
+{
+//	moduleImg = QImage(":/???"); // TODO: add icon
```
zaps166: Please ask "varlesh" for an icon :)

src/modules/Notifies/Notifies.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include <QFormLayout>
+#include <QGridLayout>
+#include <QGroupBox>
+#include <QLineEdit>
+#include <QRadioButton>
+
+ModuleSettingsWidget::ModuleSettingsWidget(Module &module) :
+	Module::SettingsWidget(module)
+{
+	/**/
```
zaps166: ?

src/modules/Notifies/Notifies.pro
```diff
@@ -0,0 +1,34 @@
...
+
+INCLUDEPATH += . ../../qmplay2/headers
+DEPENDPATH  += . ../../qmplay2/headers
+
+HEADERS += Notifies.hpp PrettyNotify.hpp TrayNotify.hpp Notify.hpp
+SOURCES += Notifies.cpp PrettyNotify.cpp TrayNotify.cpp
+
+unix:!macx:!android {
+	QT += dbus
+        HEADERS += X11_Notify.hpp
```
zaps166: tabs

src/modules/Notifies/X11_Notify.cpp
```diff
@@ -0,0 +1,146 @@
...
+		arg.beginStructure();
+		arg << 0 << 0 << 0 << false << 0 << 0 << QByteArray();
+		arg.endStructure();
+		return arg;
+	}
+	const QImage scaled = image.scaledToHeight(100, Qt::SmoothTransformation).convertToFormat(QImage::Format_ARGB32);
+#if Q_BYTE_ORDER == Q_LITTLE_ENDIAN
+	// ABGR -> ARGB
+	const QImage i = scaled.rgbSwapped();
+#else
```
zaps166: QMPlay2 doesn't support BigEndian CPUs. Do you have something (Qemu or real hardware) to test it?

src/modules/Notifies/X11_Notify.hpp
```diff
@@ -0,0 +1,44 @@
...
+*/
+
+#include <Notify.hpp>
+
+#include <QDateTime>
+
+class OrgFreedesktopNotificationsInterface;
+class QDBusPendingCallWatcher;
+
+class X11_Notify : public Notify
```
zaps166: Maybe `X11Notify`?
I prefer using camelCase instead of "_". I know that QMPlay2 has mixed styles, but new code has always `camelCase` for variables, `CamelCase` for classes name and `m_camelCase` for private/protected class members.
It can be applied also for INI (settings) keys (start e.g. from capital letter).

src/gui/Main.cpp
```diff
@@ -166,7 +166,7 @@ const QWidget *QMPlay2GUIClass::getVideoDock() const
 QMPlay2GUIClass::QMPlay2GUIClass() :
 	groupIcon(NULL), mediaIcon(NULL), folderIcon(NULL),
 	mainW(NULL),
-	screenSaver(NULL)
+	screenSaver(NULL), shortcutHandler(NULL)
```
zaps166: Unrelated change, I'll do it in `master` branch.

src/modules/Notifies/NotifyExtension.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+NotifyExtension::NotifyExtension(Module &module) : notify(NULL)
+{
+	SetModule(module);
+}
+
+bool NotifyExtension::set()
+{
+	if (sets().getBool("type_disabled"))
+		delete this->notify;
```
zaps166: I think we should avoid `this->`. Also for new code please use `m_` prefix for private/protected class members :)

src/modules/Notifies/NotifyExtension.cpp
```diff
@@ -0,0 +1,119 @@
...
+{
+	Q_ASSERT(notify);
+	if (settings.getBool("show_title"))
+		connect(&QMPlay2Core, SIGNAL(updatePlaying(bool, const QString &, const QString &, const QString &, int, bool, const QString &)), this, SLOT(updatePlaying(bool, const QString &, const QString &, const QString &, int, bool, const QString &)));
+	if (settings.getBool("show_playState"))
+		connect(&QMPlay2Core, SIGNAL(playStateChanged(const QString &)), this, SLOT(playStateChanged(const QString &)));
+	if (settings.getBool("show_volume"))
+		connect(&QMPlay2Core, SIGNAL(volumeChanged(double)), this, SLOT(volumeChanged(double)));
+
+	if(settings.getBool("custom_msg"))
```
zaps166: Always spaces after `if`, `while`, `for`, etc.

zaps166:
> Do you know about OS X?

Hmm, no :smile: Don't do for OS X (macOS), probably nobody uses QMPlay2 on macOS. It also has poor integration with it.

zaps166:
Also it is good to keep cpp and hpp in the same order including line breaks after function name, e.g. if header doesn't have line break between methods, cpp also shouldn't have it. Is it only my coding style ? :smile: ?

arthurzam:
I rebased and squashed all the commits (those renames just made my life hard when rebasing, so to fix them I rebased).

> Please ask "varlesh" for an icon :)

I think it is better that you ask him for an icon (as you are the main developer)

zaps166:
> I think it is better that you ask him for an icon (as you are the main developer)

Right :)

src/modules/Notifies/TrayNotify.cpp
```diff
@@ -0,0 +1,44 @@
...
+	this->m_timeout = timeout;
+}
+
+bool TrayNotify::showMessage(const QString &summary, const QString &message, const QString &, const QImage &)
+{
+	if (!QSystemTrayIcon::supportsMessages())
+		return false;
+
+	QSystemTrayIcon *tray = QMPlay2Core.systemTray;
+
```
zaps166:
```c++
if (!tray)
    return false;
````

**Comment by arthurzam** [comment]
Fixed.

> I can see Paused notification, but I can't see notification that playback is started again - is it intended?

Yes it is. There is no need to show "Playing" when there is a notification for the playing item (Do you have a better name for the playing item in playlist? ).

On another subject, I think it is a good idea to add a new Audio Writer for `OSS` (for `*BSD` systems). I can try to implement, but what are your thoughts?

**Comment by zaps166** [comment]
> On another subject, I think it is a good idea to add a new Audio Writer for OSS (for *BSD systems). I can try to implement, but what are your thoughts?

Who uses QMPlay2 on *BSD...? Currently they have PortAudio which it is not bad (I think) :) It makes sense if you want to use it or improve your programming skills. AFAIK OSS is very easy to implement, so you shouldn't have problems. Of course I'll review, test and merge (please open a new issue if you want to do it) :)

> There is no need to show "Playing" when there is a notification for the playing item.

Hmm, yes. But after pasued state shouldn't be shown again? Currently I can see only Paused if I'm doing sequence: pause, unpause, pause, unpause, etc...

QMPlay2 shows also 'Stopped' notification when closing and is already stopped. The same when clicking many times on stop button - maybe read/remember current state?

> Do you have a better name for the playing item in playlist?

So which name is better than "playing" ? I don't know, maybe I should ask someone who know English better than me :smile:

**Comment by arthurzam** [comment]
> Hmm, yes. But after pasued state shouldn't be shown again? Currently I can see only Paused if I'm doing sequence: pause, unpause, pause, unpause, etc...
> 
> QMPlay2 shows also 'Stopped' notification when closing and is already stopped. The same when clicking many times on stop button - maybe read/remember current state?

Done (please check the code style in this commit).

File: src/modules/Notifies/NotifyExtension.hpp [hunk 11]
```diff
@@ -33,6 +33,11 @@ private slots:
 private:
 	Notify *m_notify;
 	QString m_summaryFormat, m_bodyFormat;
+	enum PlayState{
```
zaps166: `PlayState {`

src/modules/Notifies/NotifyExtension.cpp
```diff
@@ -84,7 +85,22 @@ void NotifyService::updatePlaying(bool play, const QString &title, const QString
...
+	else if (playState == "Paused")
+		m_lastPlayState = Paused;
+	else
+		m_lastPlayState = Stopped;
+
+	/* In those cases we don't show notification:
+	 *  1. The last one is the same as the current one
+	 *  2. The current one is Playing and the last one wasn't Paused
+	 */
+	if (
```
zaps166: This line is not so long:
```c++
if (m_lastPlayState != last && (m_lastPlayState != Playing || last == Paused))
	m_notify->showMessage(QCoreApplication::applicationName(), playState);
```

arthurzam:
OK, no problem :smile:
Fixed.

src/modules/Notifies/NotifyExtension.cpp
```diff
@@ -0,0 +1,138 @@
...
+	 *  1. The last one is the same as the current one
+	 *  2. The current one is Playing and the last one wasn't Paused
+	 */
+	if (m_lastPlayState != last && (m_lastPlayState != Playing || last == Paused))
+		m_notify->showMessage(QCoreApplication::applicationName(), playState);
+}
+
+/**/
+
+NotifyExtension::NotifyExtension(Module &module) : m_notify(NULL)
```
zaps166:
```c++
NotifyExtension::NotifyExtension(Module &module) :
	m_notify(NULL)
{
```

Also maybe rename to `m_notifyService` (because not the same as in class above)?
And no need to line break between functions (consistent to header, if you have a time you can check all code for this) :D

src/modules/Notifies/NotifyExtension.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+	SetModule(module);
+}
+
+NotifyExtension::~NotifyExtension()
+{
+	delete m_notify;
```
zaps166: Open QMPlay2 when notifications are enabled. Then go to settings, disable it, apply and close QMPlay2 - crash!

src/modules/Notifies/X11Notify.cpp
```diff
@@ -0,0 +1,147 @@
...
+		arg << 0 << 0 << 0 << false << 0 << 0 << QByteArray();
+		arg.endStructure();
+		return arg;
+	}
+	const QImage scaled = image.scaledToHeight(100, Qt::SmoothTransformation).convertToFormat(QImage::Format_ARGB32);
+#if Q_BYTE_ORDER == Q_LITTLE_ENDIAN
+	// ABGR -> ARGB
+	const QImage i = scaled.rgbSwapped();
+#else
+	// ABGR -> GBAR
```
zaps166: Is it tested, does it make any sense (QMPlay2 doesn't support big endian - never tested)?

src/modules/Notifies/NotifyExtension.cpp
```diff
@@ -0,0 +1,138 @@
...
+}
+
+bool NotifyExtension::set()
+{
+	if (sets().getBool("TypeDisabled"))
+		delete m_notify;
```
zaps166: Set it to `NULL` to prevent crash described above :)

src/modules/Notifies/NotifyExtension.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include <QMPlay2Extensions.hpp>
+
+class Notify;
+
+class NotifyService : public QObject
+{
+	Q_OBJECT
+public:
+	NotifyService(Notify *m_notify, Settings &settings);
```
zaps166: `m_notify` not a class memeber here :)

src/modules/Notifies/Notify.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include <ModuleCommon.hpp>
+
+#include <QImage>
+
+class Notify : public QObject
+{
+public:
+	virtual bool showMessage(const QString &summary, const QString &message = QString(), const QString &icon = QString(), const QImage &image = QImage()) = 0;
+
+	virtual ~Notify() { }
```
zaps166: virtual ~Notify()
{}

src/modules/Notifies/Notify.hpp
```diff
@@ -0,0 +1,37 @@
...
+protected:
+	Notify(qint32 timeout) : m_timeout(timeout) {}
```
zaps166: The same as something else, new line after `:` and new line before `{}` :)

src/modules/Notifies/X11Notify.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+#include "dbus/notification.h"
+
+#include <QImage>
+
+#include <QDBusArgument>
+
+#define X11NotifyName "X11 Notify"
+
+QDBusArgument &operator<<(QDBusArgument &arg, const QImage &image)
```
zaps166: I can compile & run without this, hmm... Is it necessary?

src/modules/Notifies/X11Notify.cpp
```diff
@@ -0,0 +1,147 @@
...
+		// Reuse the existing popup if it's still open.  The reason we don't always
+		// reuse the popup is because the notification daemon on KDE4 won't re-show
+		// the bubble if it's already gone to the tray.
+		id = m_notificationId;
+	}
+
+	QDBusPendingReply<uint> reply = m_interface->Notify(
+				QCoreApplication::applicationName(), id, icon, summary,
+				message, QStringList(), hints, m_timeout);
+	QDBusPendingCallWatcher* watcher = new QDBusPendingCallWatcher(reply, this);
```
zaps166: Start to the right :D

src/modules/Notifies/X11Notify.cpp
```diff
@@ -0,0 +1,147 @@
...
+	Q_ASSERT(0); // This is needed to link but shouldn't be called.
+	return arg;
+}
+
+
+X11Notify::X11Notify(qint32 timeout) :
+	Notify(timeout),
+	m_notificationId(0)
+{
+	m_interface = new OrgFreedesktopNotificationsInterface(
```
zaps166: I think that no need to split line in QMPlay2. Splitting line using only tabs (which QMPlay2 uses) is not good :D

src/modules/Notifies/X11Notify.cpp
arthurzam: > Start to the right

Sorry, but I didn't understand.

src/modules/Notifies/X11Notify.cpp
arthurzam: > I can compile & run without this, hmm... Is it necessary?

In previous versions of Qt it was needed to convert properly `QImage` -> dbus information. For now we don't send images at all in QMPlay2 (we send only text), but for future needs.

src/modules/Notifies/X11Notify.cpp
zaps166: `QDBusPendingCallWatcher* watcher` -> `QDBusPendingCallWatcher *watcher`, but not important

zaps166:
This contains an `X11Notify` class, but it uses freedesktop notifications via D-Bus. If it can be used outside X11, so maybe rename it to `FreedesktopNotify` ?

arthurzam:
Fixed :smile:

zaps166:
:+1: Now it works much better when it reuses a notification :)

zaps166:
Thanks!
Anything else can be fixed by separate commit/PR :D

PR #62, opened 2016-10-29
Use user_agent for FFmpeg >= 3.2

arthurzam:
Since FFmpeg version 3.2 user-agent option is deprecated, and we should use user_agent instead. 
[This](https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/27714b462d1bff9e9b40fbdabb39f58e79032b81) commit made this change.


zaps166:
Thanks!

PR #69, opened 2017-01-03
Little code cleanup

arthurzam:
Some little code optimizations, readability improvement and shortening.


src/modules/FFmpeg/FFDecSW.cpp
```diff
@@ -339,8 +339,6 @@ bool FFDecSW::open(StreamInfo &streamInfo, VideoWriter *)
 		return false;
 	if (codec_ctx->codec_type == AVMEDIA_TYPE_VIDEO)
 	{
-		if (codec->capabilities & CODEC_CAP_DR1)
```
zaps166: IIRC this change will produce incorrect frames (with borders?) for FFmpeg 2.2 which (I hope) still works.

arthurzam: As I checked with FFmpeg 2.2 the CODEC_FLAG_EMU_EDGE flag was always set. Maybe I'm missing something?

arthurzam: look at this [link](https://github.com/FFmpeg/FFmpeg/blob/n2.2/libavcodec/avcodec.h#L738)

zaps166: > Maybe I'm missing something?

Yes, there is at lease one difference between FFmpeg 2.2 and 2.3 in usage of this flag, see: https://github.com/zaps166/QMPlay2/commit/d657f5ca409a621871e762c9f10e53f0de3659fa

arthurzam: I understand - now will revert this patch.

zaps166: > ... I.e. this flag is now always set.

So this looks like FFmpeg 2.2.x bug :) The best way is to compile & test :smiley:

arthurzam: I will leave this one to future cleanup pull request, after many more checks...

src/modules/FFmpeg/VAAPIWriter.hpp
```diff
@@ -27,7 +27,6 @@
 
 class VAAPIWriter : public QWidget, public VideoWriter
 {
-	Q_OBJECT
```
zaps166: This class has slots.

src/modules/FFmpeg/VDPAUWriter.hpp
```diff
@@ -31,7 +31,6 @@ struct _XDisplay;
 
 class VDPAUWriter : public QWidget, public VideoWriter
 {
-	Q_OBJECT
```
zaps166: This class has slots.

src/modules/FFmpeg/VAAPIWriter.hpp
arthurzam: Fixed

src/modules/FFmpeg/VDPAUWriter.hpp
arthurzam: Fixed

src/gui/AddressDialog.hpp
```diff
@@ -22,11 +22,12 @@
 #include <AddressBox.hpp>
 
 #include <QCheckBox>
+#include <QCoreApplication>
```
zaps166: If include order doesn't matter, please group them (if necessary, eg. QMPlay2 includes, Qt includes, C includes) and sort from the longest to the shortest (in new code) :D

In this case:
```
#include <QCoreApplication>
#include <QCheckBox>
#include <QDialog>
```

arthurzam:
> I should review it carefully to not get regressions - code cleanup can do this :)

No problem, as I agree with it.

zaps166:
Hmm, it works. Let's hope that it doesn't provide difficult-to-debug regressions :)

PR #70, opened 2017-01-05
Add XSPF playlist support

arthurzam:
This pull request implements [XSPF format](https://en.wikipedia.org/wiki/XML_Shareable_Playlist_Format) support.
For parsing and writing the XML format I used classes from QtCore, so no new dependencies.

I added a new widget, `DirsList`, to the qmplay2 lib. The reason is that it is quite convenient widget and I plan to use it in my future Music Library pull request (I started working on it). Please say your thoughts how to improve this widget and if there is need to rename it.

Tested this feature using files from Clementine music player (which mainly uses XSPF) - all worked perfectly.
Tested saving my current QMPlay2 playlist (with many nested groups and various online URLs) and then loading - also worked.

I still think the PLS format is superior, but as for importing playlists from other programs this is a great addition.


src/modules/Playlists/Playlists.cpp
```diff
@@ -27,6 +28,8 @@ Playlists::Playlists() :
 
 	init("PLS_enabled", true);
 	init("M3U_enabled", true);
+	init("XSPF_enabled", true);
+	init("XSPF_relative", true);
```
zaps166: `XSPF/Relative`, `XSPF/Dirs`, etc ?

src/qmplay2/headers/Playlist.hpp
```diff
@@ -36,6 +36,15 @@ class Playlist
 			queue(0), GID(0), parent(0)
 		{}
 
+		void clear()
```
zaps166: This can go to `cpp`, but anyway, isn't better to use `entry = Entry()` to reset variables :) ?

src/qmplay2/DirsList.cpp
```diff
@@ -0,0 +1,67 @@
...
+	connect(addB, SIGNAL(clicked()), this, SLOT(addDir()));
+	QPushButton *removeB = new QPushButton(QMPlay2Core.getIconFromTheme("list-remove"), tr("&Remove folder"));
+	connect(removeB, SIGNAL(clicked()), this, SLOT(removeDir()));
+
+	QVBoxLayout *btLayout = new QVBoxLayout;
+	btLayout->addWidget(addB);
+	btLayout->addWidget(removeB);
+	btLayout->addSpacerItem(new QSpacerItem(1, 1, QSizePolicy::Fixed, QSizePolicy::Expanding));
+
+	QHBoxLayout *layout = new QHBoxLayout(this);
```
zaps166: `layout->setMargins(3)` probably will look better

src/qmplay2/DirsList.cpp
```diff
@@ -0,0 +1,67 @@
...
+	GNU Lesser General Public License for more details.
+
+	You should have received a copy of the GNU Lesser General Public License
+	along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <DirsList.hpp>
+#include <QMPlay2Core.hpp>
+
+#include <QBoxLayout>
```
zaps166: Sort Qt includes (longest to the shortest)

src/modules/Playlists/XSPF.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <Playlist.hpp>
+
+class XSPF : public Playlist
+{
+public:
+	XSPF(const QStringList &relativeDirs, bool useRelative);
+	Entries read();
+	bool write(const Entries &);
+
+	~XSPF();
```
zaps166: I know that I placed some destructors to the bottom, but this is ugly (also my old code :smile:), so please add it under constructor and then add one line break.

src/modules/Playlists/XSPF.cpp
```diff
@@ -0,0 +1,174 @@
...
+			QString url = entry.url;
+			if (m_useRelative && url.startsWith("file://"))
+			{
+				QString path = url;
+				path.remove(0, 7);
+#ifdef Q_OS_WIN
+				if (path.startsWith('/'))
+					path.remove(0, 1);
+#endif
+				foreach (const QString &dir, m_relativeDirs)
```
zaps166: `{}`

src/modules/Playlists/Playlists.cpp
```diff
@@ -56,25 +63,46 @@ QMPLAY2_EXPORT_PLUGIN(Playlists)
 
 /**/
 
+#include <DirsList.hpp>
+
+#include <QBoxLayout>
```
zaps166:
```
#include <QGridLayout>
#include <QBoxLayout>
```

src/modules/Playlists/XSPF.cpp
```diff
@@ -0,0 +1,174 @@
...
+			inEntry = false;
+		}
+	}
+	return list;
+}
+bool XSPF::write(const Entries &list)
+{
+	QByteArray buffer;
+	QXmlStreamWriter xmlWriter(&buffer);
+	xmlWriter.setAutoFormatting(true);
```
zaps166: `xmlWriter.setAutoFormattingIndent(-1);`

Use tabs - save disk space :)

arthurzam:
I hope I fixed everything.

> Loading XSPF is ~43% slower than QMPlay2-PLS.
> Saving XSPF is ~16% slower than QMPlay2-PLS.

I guess XSPF would be slower, as we need to read an XML file (instead of easier INI file). But as it won't be the default format (PLS is better for QMPlay2), speed is less a concern.

src/modules/Playlists/XSPF.cpp
```diff
@@ -0,0 +1,176 @@
...
+					}
+				}
+			}
+			xmlWriter.writeTextElement("location", url);
+		}
+		if (!entry.name.isEmpty())
+			xmlWriter.writeTextElement("title", entry.name);
+		if (entry.length >= 0.0)
+		{
+			xmlWriter.writeTextElement("duration", QString::number((qint32)(1000 * entry.length)));
```
zaps166: `duration` is in milliseconds, so it will be enough. `QMPlay_length` is not necessary in this case. Use `duration` and divide it by `1000.0` in read method.

src/modules/Playlists/XSPF.cpp
```diff
@@ -0,0 +1,176 @@
...
+			xmlWriter.writeTextElement("QMPlay_length", QString::number(entry.length, 'g', 13));
+		}
+		if (entry.selected)
+			xmlWriter.writeTextElement("QMPlay_sel", QString::number(entry.selected));
```
zaps166: `entry.selected` is always true here, just write `true` or `1` (`1` is shorter).

You can also refactor my old names, use CamelCase and QMPlay2, e.g. `QMPlay2GID`, `QMPlay2Queue`, `QMPlay2Parent`, `QMPlay2Selected` or simply drop `QMPlay2` prefix if it won't cause conflicts, e.g. `selected`, `queue`, `group`, `parent`

arthurzam:
> Also QMPlay2 has groups, maybe it should not be as <track>, but as <group id='1'> (id can be placed here instead of QMPlay_GID) ? I hope that other players will ignore this :)

It is quite unfortunate, but kaffeine will break reading XSPF if we use nested tags like you suggested.

All else were done.

src/modules/Playlists/XSPF.cpp
```diff
@@ -0,0 +1,154 @@
...
+			else if (inEntry)
+			{
+				const QString value = xmlReader.readElementText();
+				if (name == "location")
+				{
+					if (value.startsWith("file://")) // local absolute path
+						curr.url = Functions::Url(value);
+					else if (Functions::splitPrefixAndUrlIfHasPluginPrefix(value, NULL, NULL, NULL)) // plugin URL
+						curr.url = value;
+					else if(QFileInfo::exists(readFilePath + value)) // relative path
```
zaps166: space :)

src/modules/Playlists/XSPF.cpp
```diff
@@ -0,0 +1,154 @@
...
+						curr.url = Functions::Url(value, readFilePath);
+				}
+				else if (name == "title")
+					curr.name = value;
+				else if (name == "duration")
+					curr.length = value.toInt() / 1000.0;
+				else if (name == "selected")
```
zaps166: "selected" is now obsolete. XSPF shouldn't have this key, because there is no need for compatibility :)

arthurzam:
I made full rebase and squash (the commit is much cleaner now).

src/modules/Playlists/XSPF.cpp
```diff
@@ -0,0 +1,151 @@
...
+				}
+				curr = Entry();
+				inEntry = true;
+			}
+			else if (inEntry)
+			{
+				const QString value = xmlReader.readElementText();
+				if (name == "location")
+				{
+					if (value.startsWith("file://")) // local absolute path
```
zaps166: Does `Functions::fileName()` work here?

src/modules/Playlists/XSPF.cpp
```diff
@@ -0,0 +1,151 @@
+/*
+	QMPlay2 is a video and audio player.
+	Copyright (C) 2010-2016  Błażej Szczygieł
```
zaps166: 2017

src/modules/Playlists/XSPF.hpp
```diff
@@ -0,0 +1,32 @@
+/*
+	QMPlay2 is a video and audio player.
+	Copyright (C) 2010-2016  Błażej Szczygieł
```
zaps166: 2017

zaps166:
> I will think about the <extension> deal!

Ok!

Maybe relative paths should be also introduced to PLS and M3U? It will be good to achieve this from save dialog, but is it possible with native file dialog called from Qt?

zaps166:
Please also append `xspf` extension here: https://github.com/zaps166/QMPlay2/blob/f0e99a941c640a6a7d341f38f811e8b3e3296eb2/src/gui/macOS/BundleInfo.plist.in#L103 :)

arthurzam:
After various checks, I found out that `vlc` doesn't support external extensions (meaning extensions that aren't part of `vlc`). Thats why I think we should not support `vlc` (as we don't want to lose QMPlay2's information).

> Maybe relative paths should be also introduced to PLS and M3U?
> Please also append `xspf` extension here ...

Done

> It will be good to achieve this from save dialog, but is it possible with native file dialog called from Qt?

Didn't understand what you mean. Does the new patch work for you?

zaps166:
Done as ab0082f9e354afae5ef23dd18a6e6a1de9c336b6, thanks!

zaps166:
IMHO relative paths should be guessed automatically :)

zaps166:
> IMHO relative paths should be guessed automatically :)

I thought about this: https://github.com/zaps166/QMPlay2/commit/ca140e69d0df2f620b097e6adea7a0bd30fdc2d2

PR #78, opened 2017-01-13
Add Skip & Stop After track

arthurzam:
This pull request adds "skip track" and "stop after track" to playlist.

Those flags are saved using OR between flags and saving the flags as int instead of selected (in PLS: newer version reads without problem old format, while older version reads newer format without selected field).

Skip track makes, that when the playlist automagically moves to the skipped track, it moves to the next one. Shown using strike through the title text.

Stop after track is to stop when the selected track finishes playing, if the flag is set then it just stops the play. Shown using the "stop" string near the time.

Please say your thoughts about the way it is shown to the user.

### DIFF
--- a/src/gui/MenuBar.cpp
+++ b/src/gui/MenuBar.cpp
@@ -155,6 +155,8 @@ MenuBar::Playlist::Playlist(MenuBar *parent) :
 	newAction(Playlist::tr("&Go to the playback"), this, goToPlayback, false, QIcon(), false);
 	addSeparator();
 	newAction(Playlist::tr("&Enqueue"), this, queue, false, QIcon(), false);
+	newAction(Player::tr("&Skip"), this, skip, true, QIcon(), false);
+	newAction(Player::tr("&Stop after this track"), this, stopAfter, true, QIcon(), false);
 	addSeparator();
 	newAction(Playlist::tr("&Properties"), this, entryProperties, false, QMPlay2Core.getIconFromTheme("document-properties"), false);
 }
@@ -458,6 +460,8 @@ void MenuBar::setKeyShortcuts()
 	shortcuts->appendAction(playlist->expandAll, "KeyBindings/Playlist-expandAll", "");
 	shortcuts->appendAction(playlist->goToPlayback, "KeyBindings/Playlist-goToPlayback", "Ctrl+P");
 	shortcuts->appendAction(playlist->queue, "KeyBindings/Playlist-queue", "Q");
+	shortcuts->appendAction(playlist->skip, "KeyBindings/Playlist-skip", "");
+	shortcuts->appendAction(playlist->stopAfter, "KeyBindings/Playlist-stopAfter", "");
 	shortcuts->appendAction(playlist->entryProperties, "KeyBindings/Playlist-entryProperties", "Alt+Return");
 
 	shortcuts->appendAction(playlist->add->file, "KeyBindings/Playlist-Add-file", "Ctrl+I");
--- a/src/gui/PlaylistDock.cpp
+++ b/src/gui/PlaylistDock.cpp
@@ -111,7 +111,9 @@ bool PlaylistDock::save(const QString &_url, bool saveCurrentGroup)
 		entry.name = tWI->text(0);
 		if (tWI->parent())
 			entry.parent = parents.indexOf(tWI->parent()) + 1;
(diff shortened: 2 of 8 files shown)


src/qmplay2/headers/Playlist.hpp
```diff
@@ -30,15 +30,21 @@ class Playlist
...
 			length(-1.0),
-			selected(false),
-			queue(0), GID(0), parent(0)
+			flags(0), queue(0), GID(0), parent(0)
 		{}
 
 		QString name, url;
 		double length;
-		bool selected;
+		int flags;
```
zaps166: `qint32` for consistency; you can squash three variables to one line.

src/gui/MenuBar.cpp
```diff
@@ -458,6 +460,8 @@ void MenuBar::setKeyShortcuts()
 	shortcuts->appendAction(playlist->expandAll, "KeyBindings/Playlist-expandAll", "");
 	shortcuts->appendAction(playlist->goToPlayback, "KeyBindings/Playlist-goToPlayback", "Ctrl+P");
 	shortcuts->appendAction(playlist->queue, "KeyBindings/Playlist-queue", "Q");
+	shortcuts->appendAction(playlist->skip, "KeyBindings/Playlist-skip", "");
```
zaps166: Maybe add default shortcuts?

src/gui/PlaylistDock.cpp
```diff
@@ -104,14 +104,19 @@ bool PlaylistDock::save(const QString &_url, bool saveCurrentGroup)
...
-			entry.queue = tWI->text(1).toInt();
+			entry.length = tWI->data(3, Qt::UserRole).isNull() ? -1.0 : tWI->data(3, Qt::UserRole).toDouble(); //FIXME: Is it possible to have a null variant for non-groups entries?
+			entry.queue = tWI->text(2).toInt();
 		}
 		entry.url = tWI->data(0, Qt::UserRole).toString();
 		entry.name = tWI->text(0);
 		if (tWI->parent())
 			entry.parent = parents.indexOf(tWI->parent()) + 1;
-		entry.selected = tWI == list->currentItem();
+		if(tWI == list->currentItem())
```
zaps166: Spaces :smile:

src/gui/PlaylistDock.cpp
```diff
@@ -285,30 +297,32 @@ void PlaylistDock::next(bool playingError)
...
 							break;
 						}
-						if (PlaylistWidget::isGroup(tWI)) //grupa
+						if (PlaylistWidget::isGroup(tWI)) //group
 						{
 							if (!tWI->isExpanded())
 								tWI->setExpanded(true);
 							continue;
 						}
+						if (tWI->font(0).strikeOut()) //skip item
```
zaps166: Is it a good idea to check font attributes? Maybe a flag in user data or something?

src/gui/PlaylistWidget.cpp
```diff
@@ -660,10 +660,18 @@ QTreeWidgetItem *PlaylistWidget::newEntry(const Playlist::Entry &entry, QTreeWid
...
+	{
+		QFont font = tWI->font(0);
+		font.setStrikeOut(true);
+		tWI->setFont(0, font);
+	}
 	tWI->setText(0, entry.name);
 	tWI->setData(0, Qt::UserRole, entry.url);
-	tWI->setText(2, Functions::timeToStr(entry.length));
-	tWI->setData(2, Qt::UserRole, entry.length);
+	tWI->setText(1, ((entry.flags & Playlist::Entry::FlagStopAfter) ? tr("stop") : ""));
```
zaps166: Playlist shouldn't have additional column - looks bad to me. Maybe use italic font for stop-after entry? It'll also simplify everything (less changes, etc) :)

src/gui/MenuBar.cpp
arthurzam: Which one can be used? (I don't know which to choose)

arthurzam:
I fixed all your reviews, except of default shortcuts 😃

zaps166:
Oops :smile: 

> If you want to store more then one custom value in the same column, you can use e.g. Qt::UserRole+1 or use a global constant for it.

Should be `Qt::UserRole+2`: 3f109ee94fdf37c510763e3c2ee9d12d60161fc3

PR #82, opened 2017-01-27
Change menu's separator color to border's color

arthurzam:
This small pull request changes menu's separator color to be the same as border's color. For example during QMPlay2 dark theme, the separator wasn't seen at all.

What are your thoughts to adding a multiline text box in the bottom of appearance dialog for setting the stylesheet of the app (good for advanced users)?

### DIFF
--- a/src/gui/Appearance.cpp
+++ b/src/gui/Appearance.cpp
@@ -77,7 +77,7 @@ static const QString QMPlay2ColorExtension = ".QMPlay2Color";
 static QPalette systemPalette;
 static QString colorsDir;
 
-static void setBrush(QPalette &pal, QPalette::ColorRole colorRole, QColor color, double alpha = 1.0)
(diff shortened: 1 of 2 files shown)


zaps166:
Now the menu separator is too thick (depends on qt style). Also CSS causes other problems (e.g. go to full screen and open context menu - it is empty depends on qt style, etc.). Now the CSS engine runs for all widgets.

> For example during QMPlay2 dark theme, the separator wasn't seen at all.

It is visible, but poorly (#0A0811 vs #08060D on Fusion style) :)

> What are your thoughts to adding a multiline text box in the bottom of appearance dialog for setting the stylesheet of the app (good for advanced users)?

CSS + QPalette doesn't work good in all cases, so CSS **or** QPalette should be used. I think it is not necessary now, but maybe in a future :)

PR #83, opened 2017-01-28
Add support for multiple playlists

arthurzam:
This is a middle sized pull request for adding support for multiple playlists for QMPlay2. This includes:

- All playlists are stored in `~/.qmplay2/Playlists/${number}_${name}.pls` (when number is for ordering the playlists and the name of the playlist). Original `~/.qmplay2/Playlist.pls` is loaded as "Main" first playlist and then the file is deleted.
- Added actions to "Playlist" menu for creating new playlist (named by default as "Playlist"), delete the current playlist and rename current playlist. Original action for load and save playlist were changed for saving current playlist and load playlist to a new playlist. I set some default shortcuts for those actions.
- Renamed private variables in `PlaylistDock` for the `m_` prefix.
- The playlists are shown using a `QTabWidget`, in which every tab has a `PlaylistWidget` widget. A button at start of tab row is for adding a new playlist. Tabs are movable (as a result the order of playlists is changed). On double click on tab name is possible for the user to rename the playlist. It is also possible to close the tab.
- Tested play, pause, stop, next, prev, edit properties and etc - all worked for me.

Work to be done for the future (outside of this pull request in my idea):

1. I found a bug that when we have at least 2 playlists, the first one is playing and we double click on item from the second playlist, the icon of the previous playing item doesn't change to original. I didn't manage to fix it. We need to change something in `PlaylistDock::itemDoubleClicked`.
2. Add an option for making a playlist as read only (non renaming and non changing order, non adding/removing items and etc). For this we need to add an "lock" flag to `PlaylistWidget`. But how to save those flags?
3. Add support for Drag and Drop between tabs (by hovering the tab name during the drag).

This pull request is a start for some great features, enabling multiple playlists. I think feature wise we can merge it (of course after code review) and then continue to fix and add features (like those mentioned above).

### DIFF
--- a/src/gui/MainWidget.cpp
+++ b/src/gui/MainWidget.cpp
@@ -782,6 +782,7 @@ void MainWidget::createMenuBar()
 	connect(menuBar->playlist->loadPlist, SIGNAL(triggered()), this, SLOT(loadPlist()));
 	connect(menuBar->playlist->savePlist, SIGNAL(triggered()), this, SLOT(savePlist()));
 	connect(menuBar->playlist->delPlist, SIGNAL(triggered()), playlistDock, SLOT(playlistsCloseTab()));
+	connect(menuBar->playlist->renamePlist, SIGNAL(triggered()), playlistDock, SLOT(playlistsTabDoubleClicked()));
 	connect(menuBar->playlist->saveGroup, SIGNAL(triggered()), this, SLOT(saveGroup()));
 	connect(menuBar->playlist->newGroup, SIGNAL(triggered()), playlistDock, SLOT(newGroup()));
 	connect(menuBar->playlist->renameGroup, SIGNAL(triggered()), playlistDock, SLOT(renameGroup()));
--- a/src/gui/MainWidget.cpp
+++ b/src/gui/MainWidget.cpp
@@ -385,7 +385,7 @@ MainWidget::MainWidget(QList<QPair<QString, QString>> &arguments) :
 		}
 	}
 
-	playlistDock->load(QMPlay2Core.getSettingsDir() + "Playlist.pls");
+	playlistDock->loadAll();
 
 	bool noplay = false;
 	for (const auto &argument : asConst(arguments))
@@ -1708,7 +1708,7 @@ void MainWidget::closeEvent(QCloseEvent *e)
 
 	if (settings.getBool("AutoDelNonGroupEntries"))
 		playlistDock->delNonGroupEntries(true);
-	playlistDock->save(QMPlay2Core.getSettingsDir() + "Playlist.pls");
+	playlistDock->saveAll();
 
 	playC.stop(true);
 }
--- a/src/gui/MainWidget.cpp
+++ b/src/gui/MainWidget.cpp
@@ -1201,7 +1201,7 @@ void MainWidget::loadPlist()
 }
 void MainWidget::savePlist()
 {
-	savePlistHelper(tr("Save playlist"), QMPlay2GUI.getCurrentPth(playlistDock->getUrl()), false);
+	savePlistHelper(tr("Save playlist \"%1\"").arg(playlistDock->getCurrentPlaylistName()), QMPlay2GUI.getCurrentPth(playlistDock->getUrl()), false);
 }
 void MainWidget::saveGroup()
 {
--- a/src/gui/MainWidget.cpp
+++ b/src/gui/MainWidget.cpp
@@ -780,6 +780,7 @@ void MainWidget::createMenuBar()
 	connect(menuBar->playlist->quickSync, SIGNAL(triggered()), playlistDock, SLOT(quickSyncCurrentFolder()));
 	connect(menuBar->playlist->loadPlist, SIGNAL(triggered()), this, SLOT(loadPlist()));
 	connect(menuBar->playlist->savePlist, SIGNAL(triggered()), this, SLOT(savePlist()));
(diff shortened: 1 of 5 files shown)


zaps166:
Thanks! This is a design change, so I'll defer it a little bit.

---

> 3. Add support for Drag and Drop between tabs (by hovering the tab name during the drag).

Is it #84, 3e7a63953ce415bde7ed8b8699d3c2f22d769674?

---

In this PR please split for more commits:
- move my code to other file,
- change the old code (e.g. variable names if necessary or bug fixes),
- add a new feature trying to not modify what is not necessary for the new feature.

Changing variables name for unrelated changes (e.g. add `m_` prefix) expands the diff, so it is difficult to read important changes.

arthurzam:
> Is it #84, 3e7a639?

Not really. I mean drag and drop of `QTreeWidgetItem` between tabs.

> move my code to other file,
> change the old code (e.g. variable names if necessary or bug fixes),

What does you mean?

> add a new feature trying to not modify what is not necessary for the new feature

The movement of some functions from `MainWidget` to `PlaylistDock` was for more encapsulated code.

> Changing variables name for unrelated changes (e.g. add m_ prefix) expands the diff, so it is difficult to read important changes.

I understand. After I understand what you meant with previous points, I will change appropriately the pull request.

zaps166:
> Not really. I mean drag and drop of QTreeWidgetItem between tabs.

I didn't test yet, but it is very similar. The dragged item should switch the tab if it is over it, right?

> The movement of some functions from MainWidget to PlaylistDock was for more encapsulated code.

Is it necessary for a feature?

> What does you mean?

Split it for more commits (move existing code, change variable names, add a feature, etc). But it is better to only add something and modify only what is necessary - without unneeded changes (i.e. refactoring, etc). Too many changes means too difficult to see the feature in the code :) Especially that I want to refactor playlist in the future :)

---

The playlist code (PlaylistWidget and PlaylistDock) is ugly, but it works quite good, so no need for unnecessary changes... I'll refactor it or rewrite in the future with new features :)

arthurzam:
I rebased the code and split for feature divided commits. Removed unneeded variables rename (but my new variables are prefixed with `m_`). 

I managed to implement without moving the 3 functions from `MainWidget` to `PlaylistDock`, but I think it looks uglier.

arthurzam:
I rebased (as there were merge conflicts). Waiting for your review 😄

zaps166:
👍 Thanks!
Deferred, I'll review/test, but later :)

arthurzam:
Rebased code to fix merge conflict.
Also ported the code to `C++11` (like changing `foreach` to `for` and `NULL` to `nullptr`).
No other changes were made 😄

zaps166:
Just tested quickly - it doesn't save the playlist. Also it shouldn't display tabs if only one playlist exists :)

arthurzam:
> Just tested quickly - it doesn't save the playlist

Can you please say the actions list? Because I test it with various actions and it saves OK.

> Also it shouldn't display tabs if only one playlist exists

Done

arthurzam:
By the way, another more global question:
After the move to `C++11` we have more options for optimize or improve code readability.
One of those options is for simple types (like int. pointers) to set their's value in the header file inline with their's declaration, and remove from constructor definition. This change will reduce code lines, and improve seeing the default values. There are a lot of places for this change. What do you think about this change (I can do it in a different PR).

Another possible improvement for performance is to add in some places another function version receiving a `std::initializer_list<QString>` for example and in the function to traverse this initializer list.

In other places we can change `QString` to `QLatinString` (no change for Qt4 but greatly improves for Qt5) in places where we use internal strings which won't be translated at all.

What are your thoughts about those options? It can improve in many places the readability and performance. Of course not in this PR, but in a new one.

zaps166:
> Can you please say the actions list? Because I test it with various actions and it saves OK.

IIRC: Open QMPlay2 - loads Main playlist, then close and open again - there is no playlist, no PLS file. Add something, close and open again - empty playlist, no PLS file.

zaps166:
> One of those options is for simple types (like int. pointers) to set their's value in the header file inline with their's declaration, and remove from constructor definition.

Good idea, but do you want to fix every class ? :) But if you'll do it - please do it only for default values (zero). If it will have other value - initialize it in constructor in initialization list (except simple structures which doesn't have explicit constructor).

> Another possible improvement for performance is to add in some places another function version receiving a std::initializer_list<QString>

I already use it in Qt/STL containers, but do we need it in QMPlay2 classes? Please point me where it can be useful :)

> In other places we can change QString to QLatinString

`QLatin1String` - it prevents Utf-8/Utf-16, but is it faster than `QByteArray`? Please notice that QMPlay2 already uses `QStringBuilder` which can reduce memory allocations :)

---

Please move your message about C++11 to new Issue. Then I'll move this and we'll remove comments about C++11 from here :smile:

arthurzam:
Well, now I finally returned home and can work again on this PR 😄 

> Open QMPlay2 - loads Main playlist, then close and open again - there is no playlist, no PLS file. Add something, close and open again - empty playlist, no PLS file.

Please check what happens with your `.qmplay` settings directory. I ran firstly the `master` branch (without multiple playlists), and than applied this patch series and tested those actions - everything works excellently!

Please note, that when you close all tabs, an empty playlist is created (thus there is always at least one playlist ready). Maybe you misunderstood because of it.

zaps166:
This PR still has a lot of bugs in UI and logic: too large margins, corrupted entries in context menu, doesn't store the playlist(s) at exit (removes it), etc.

zaps166:
Closing due to long time inactivity.

PR #89, opened 2017-02-18
Add SoundCloud support

arthurzam:
This pull request adds SoundCloud support. This includes:

1. `USE_SOUNDCLOUD` flag
2. The icon is downloaded from official SoundCloud. I hope this icon is OK for now.
3. `client_id` (API key of app) is taken from Clementine (I didn't manage to create a new API key). If you manage to create a new one, I can replace it.
4. Supports searching tracks (UI the same as ProstoPleer).
5. Supports copying regular browser URLs from SoundCloud tracks.


zaps166:
Very nice extension, thanks! :)

Bugs:
- when I want to find something with space in name it shows a connection error,
- title column is the most important, but it is too small because of Genre, please remove the line with `ResizeToContents` from Genre column,
- I hope that Clementine developers allows to use their ID 😄,
- `QJsonDocument` doesn't exist in Qt4 - also please don't use external libs, so enable it only for Qt5 by default in CMake and QMake (don't write your own parser ATM (it is task for me) 😄).

---

> Please be aware that we transcode all tracks to 128kbps mp3 for streaming playback. When you make your track downloadable, however, this allows your listeners to download your track in the same format you uploaded it in, without any additional transcoding.

It doesn't sound good 😢

zaps166:
Also missing completer in combo box - if it is not possible or difficult - remove `QCompleter` from includes :)

arthurzam:
> when I want to find something with space in name it shows a connection error,
> title column is the most important, but it is too small because of Genre, please remove the line with ResizeToContents from Genre column,
> QJsonDocument doesn't exist in Qt4 - also please don't use external libs, so enable it only for Qt5 by default in CMake and QMake (don't write your own parser ATM (it is task for me) 😄).

Done. For now this use flag is enabled only if Qt5 is detected.

> Also missing completer in combo box

 Could not find how to interact with API to get this, so cleaned includes.

> I hope that Clementine developers allows to use their ID

Well, the API key was published on the code itself. I took some inspiration how to implement it without `OAuth2` (but I think for better support for various online streams we will have to use it in the future. Maybe [o2 lib](https://github.com/pipacs/o2) is good for it). On the other hand, we use the API fairly, so as owner of QMPlay2, you can register the app at `SoundCloud` to get our own API key.

> It doesn't sound good 😢

Never knew about it. But bad quality support is better than no support in QMPlay2. Maybe inform in the README file about this issue? Or even better, as the README file is to big already, add information to wiki pages in GitHub (see in repo's settings if you are interested).

zaps166:
> so as owner of QMPlay2, you can register the app at SoundCloud to get our own API key.

I've started and then resigned - too many things to do for app id. Let's use a public id.

zaps166:
> Maybe inform in the README file about this issue?

No need for this - it is a soundcloud limitation, not QMPlay2 limitation.

zaps166:
Oops, still wants to compile it on Qt4 by default

zaps166:
> Oops, still wants to compile it on Qt4 by default

QtCreator cached it, sorry 😄

zaps166:
> README file is to big already

Yeah, readme it is very long 😄 I'll create wiki in the future...

zaps166:
@arthurzam SoundCloud no longer works, is it api key issue or something else is wrong ?

PR #97, opened 2017-04-15
fix deprecated cmake function

arthurzam:
PRINT_ENABLED_FEATURES and PRINT_DISABLED_FEATURES are deprecated since CMake 3.8.
- Checked using CMake 2.8


zaps166:
Thanks!

PR #107, opened 2017-05-09
use QByteArray::constData() instead of QByteArray::data()

RJVB:
Common practice for read-only access; the Qt documentation claims this is (can be) faster because no data is ever (deep) copied.
The patch modifies the locations where the potential gain should be the most beneficial.
There is 1 other location PortAudioWriter::writeStream(); that hunk intersects with another patch I plan to propose and will thus be part of the corresponding PR (which doesn't mean it cannot be submitted along with these changes).

### DIFF
--- a/src/modules/Inputs/PCM.cpp
+++ b/src/modules/Inputs/PCM.cpp
@@ -91,7 +91,7 @@ bool PCM::read(Packet &decoded, int &idx)
 	const int samples_with_channels = dataBA.size() / bytes[fmt];
 	decoded.resize(samples_with_channels * sizeof(float));
 	float *decoded_data = (float *)decoded.data();
-	ByteArray data(dataBA.data(), dataBA.size(), bigEndian);
+	ByteArray data(dataBA.constData(), dataBA.size(), bigEndian);
 	switch (fmt)
 	{
 		case PCM_U8:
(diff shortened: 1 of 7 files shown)


zaps166:
Every `QByteArray` variable which is changed in this PR is already `const`, so `data()` and `constData()` does the same thing (doesn't detach). The only thing which I'm not sure is `line.toLatin1().constData()` and `time[i].toLatin1().constData()` - maybe this is a good change :)

RJVB:
On Tuesday May 09 2017 10:59:25 Błażej Szczygieł wrote:
> Every `QByteArray` variable which is changed in this PR is already `const`, so `data()` and `constData()` does the same thing (doesn't detach). 

Hmm, if you say so. I wouldn't like to rely on such indirect control and prefer to see explicitly when I'm getting r/o access and when r/w. Still a C programmer at heart I guess.

I don't see any difference in the functions called by `data()` and `constData()` beyond the `const` keyword in the latter. So apparently it's the compiler's decision to make a "deep copy" or not, and can that even happen when you just do

`char *c = arr.data();`

and then modify the contents of `c` with the intent of modifying `arr`?


> The only thing which I'm not sure is `line.toLatin1().constData()` and `time[i].toLatin1().constData()` - maybe this is a good change :)

In my opinino they're all good chances because you won't get errors (compiler or otherwise) when you start copy/pasting code :)

R

zaps166:
> In my opinino they're all good chances

Right, I can use `constData()` everywhere :) But your PR doesn't contain all occurrences of `QByteArray::data()` :(

RJVB:
No, I haven't been keeping this up-to-date after I realised that most if not all my suggested improvements were moot.

Issue #105, opened 2017-05-07
[Feature Request] output device sampling rate control for bit-perfect audio


RJVB:
This is a bit of a pet horse for me: getting bit-perfect audio output. This is an approach where you send the demuxed and decodec audio stream "as is" to a DAC like a headphone amp or an AV amplifier. IOW, you adapt the output sampling rate of your audio device to the content instead of resampling the content to match the current device sampling rate, with the goal of getting as pure a signal as possible into the best DAC you might have. Evidently there is also a (small) cost to sampling rate conversion so skipping that step frees the CPU for other tasks, might improve audio/video synchronisation, etc.

I'm not certain to what extent this is possible on Linux desktops where PulseAudio is running but it certainly should be possible on MSWin (using WASAPI) and Mac.

This is something I'd probably enjoy contributing myself, for Mac. As far as I can tell you only support PortAudio on that platform; do you happen to know if PortAudio has an API for setting the device output rate? That might be easiest, but I have already written a plugin for iTunes once that does the same thing so if could contribute some "native" code too if required.

zaps166:
> This is an approach where you send the demuxed and decodec audio stream "as is" to a DAC like a headphone amp or an AV amplifier.

I don't understand... :)

---

QMPlay2 demuxes, decodes and sends samples as-is to the output (PortAudio in your case). The only thing which QMPlay2 does is converting samples to 32-bit `float` (if decoded audio is not `float`). If the output (PortAudio->CoreAudio, ALSA, PulseAudio, OpenSLES, etc.) is unable to open desired sample rate (song sample rate), QMPlay2 tries to use supported sample rate and it resamples the audio. You can also go to Playback Settings and force the sample rate manually - then QMPlay2 will also resample.

On macOS you can set CoreAudio sample rate [here](https://support.apple.com/library/content/dam/edam/applecare/images/en_US/imac/audio_midi_setup.png) - adjust it to your song sample rate if DAC properly supports this sample rate (e.g. on my PC motherboard, 44.1 kHz doesn't sound good, but 48 kHz is good, so it is better to use high quality resampler in OS, e.g. `speex-float-5` in PulseAudio).

FFmpeg resampler is used in QMPlay2 (libswresample) with linear interpolation enabled: https://github.com/zaps166/QMPlay2/blob/master/src/qmplay2/SndResampler.cpp

RJVB:
On Sunday May 07 2017 03:18:29 Błażej Szczygieł wrote:
[...]

PR #108, opened 2017-05-09
Allow bit-perfect audio output where supported

RJVB:
When QMPlay2 isn't configured to resample all audio content to a user-specific sample rate it sends the demuxed and decoded audio stream "as is" to the selected audio device. When using the PortAudio module a stream is opened that is set to the content sample rate, but no attempt is made to configure the device appropriately.
This patch introduces an initial implementation of a bit-perfect audio option in the PortAudio module. After opening the PA stream, platform-native APIs are called to switch the selected audio device to the content's sample rate or to the most appropriate (closest) supported sample rate.
The initial device settings are cached and restored when playback is stopped.

The goal is to create the optimum playback conditions where the device will not introduce unnecessary aliasing artefacts. Evidently this cannot catch all non-ideal situations: content with higher sample rate than the device can handle (say 192kHz vs. 48kHz) will sound sub-optimally (or just plain bad). Such cases would require resampling down (an integer factor) to the most appropriate supported sample rate and set the device to that rate. I don't see how to force the resample step from within an output module, though.

The patch also
- contains a QByteArray optimisation
- checks for changes to the default PA device before opening the PA stream


zaps166:
Isn't better to write the entire CoreAudio module?

RJVB:
On Thursday May 11 2017 07:48:17 Błażej Szczygieł wrote:
>Isn't better to write the entire CoreAudio module?

Bypassing PortAudio, you mean?

Yes, I guess it would be. One less dependency, and also PortAudio apparently doesn't detect when you change the output device while QMPlay2 is running.

But I'll admit that I don't really have the time to start figuring out how to write such a module properly.

Another argument against: with the current approach porting the new feature to other platforms only requires porting the AudioDevice classes. I'm thinking about MSWin in particular: most PCs have HDMI ports nowadays, and MS Windows can do audio-only over HDMI.

zaps166:
> But I'll admit that I don't really have the time to start figuring out how to write such a module properly.

Maybe I'll try to do something with CoreAudio if macOS version is useful for other people :) I know that PortAudio might not be sufficient... I'll ask you about macOS things when I start writing it, OK?

> in particular: most PCs have HDMI ports nowadays, and MS Windows can do audio-only over HDMI.

What do you mean? What about internal notebook speakers, analog Jack output, DisplayPort, USB devices, Bluetooth devices? 😄 

---

Current Windows PortAudio implementation doesn't detect audio device changes w/o restarting QMPlay2, but it can play via new audio device even without restarting playback (e.g. play via internal speakers, connect HDMI and it can automatically continue playing via HDMI).

RJVB:
On Thursday May 11 2017 08:18:57 Błażej Szczygieł wrote:

> > But I'll admit that I don't really have the time to start figuring out how to write such a module properly.
> 
> Maybe I'll try to do something with CoreAudio if macOS version is useful for other people :) I know that PortAudio might not be sufficient... I'll ask you about macOS things when I start writing it, OK?

Fine, and I'll try to help!

> > in particular: most PCs have HDMI ports nowadays, and MS Windows can do audio-only over HDMI.
> 
> What do you mean? What about internal notebook speakers, analog Jack output, DisplayPort, USB devices, Bluetooth devices? 😄 

Yeah, I realised my argument wasn't really complete :)
What I mean is that HDMI gives you a cheap and easily available option to use a PC as a content source with the quality DAC in an AV ("home cinema") amplifier . Often the entry-level models have the same components that matter as the more expensive models with all kinds of bells and whistles. The one I have didn't cost more than a good USB headphone amplifier, except that it can drive up to 5 big speakers, has optic inputs, a built-in radio etc. It's a prime candidate for bit-perfect audio output, but not any more so than other external devices ... or an internal soundcard if it has a sufficiently shielded line-out. The analog out from most PCs (and even Macs) rarely puts out a quality signal so aiming for the highest possible quality doesn't make much sense with that output.
DisplayPort on PCs? I suppose it exists, but I've yet to encounter it in laptops and notebooks that cost less than a Mac ;)


> Current Windows PortAudio implementation doesn't detect audio device changes w/o restarting QMPlay2

Maybe that's a cross-platform design flaw then?

> , but it can play via new audio device even without restarting playback (e.g. play via internal speakers, connect HDMI and it can automatically continue playing via HDMI).

That doesn't work on Mac.

R.

zaps166:
> DisplayPort on PCs?

Most modern nvidia graphics cards have DisplayPort (including my graphics card), see (Polish): https://www.x-kom.pl/g-5/c/346-karty-graficzne-pci-e.html?f[1710][73038]=1 :)

> That doesn't work on Mac.

Yes, because that note was about Windows.

RJVB:
On Thursday May 11 2017 10:00:07 Błażej Szczygieł wrote:
> Most modern nvidia graphics cards have DisplayPort (including my graphics card), see (Polish): https://www.x-kom.pl/g-5/c/346-karty-graficzne-pci-e.html?f[1710][73038]=1 :)

Good to know, but are those available to the OS as if they were on the motherboard?
Still doesn't help (potential) notebook owners, though ;)

> Yes, because that note was about Windows.

Yes, and I just confirmed that it only works on MS Windows O:-)

zaps166:
IMHO it should be done as a CoreAudio module instead of modifying PortAudio module - too much macOS code. E.g. Linux has native ALSA and PulseAudio support.

RJVB:
That, or it should be upstreamed into PortAudio where all client software gets to benefit from it. The sample rate parameter is already being passed in, just as well use it :)

I've proposed my patch upstream, but the project uses a repo of a paying service so I'm not even certain they ever got my proposition (before my evaluation account expired).

zaps166:
Closing, see #105

src/modules/PortAudio/Mac/AudioDevice.h
```diff
@@ -0,0 +1,136 @@
+/*
```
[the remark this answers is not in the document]
RJVB: I don't see why it couldn't. The license boilerplate even suggests that it doesn't *have* to be preserved for when you modify the software and/or redistribute only parts of it. We should be fine  when we keep it.

Either way, this comes from example code showing how certain things need to be done. It's not like there are any number of other ways to do the same thing.

src/modules/PortAudio/PortAudioWriter.cpp
```diff
@@ -191,7 +204,30 @@ void PortAudioWriter::pause()
 
 QString PortAudioWriter::name() const
 {
-	return PortAudioWriterName;
```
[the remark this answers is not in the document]
RJVB: This I don't get. Why not provide the more detailed information everywhere? I propose to add the content sample rate only on Mac (aka platforms where bitperfect audio is supported, if ever there will be more) but NOT to check for the bitperfect flag. This will tell the user at any time how sample rate is being handled during playback: conversion by CoreAudio (bitperfect OFF) or by the hardware device.

RJVB:
I'm kind of allergic to the 'macOS' naming which is inappropriate for older OS X and Mac OS X versions - I'll use the more specific and `3rdparty/CoreAudio` path (more generic too, since there is a CoreAudio implementation for MS Windows ;) ).

src/modules/PortAudio/PortAudio.cpp
```diff
@@ -88,15 +88,23 @@ ModuleSettingsWidget::ModuleSettingsWidget(Module &module) :
...
+	bitPerfect = new QCheckBox(tr("Bit-perfect audio"));
+	bitPerfect->setChecked(sets().getBool("BitPerfect"));
+	bitPerfect->setToolTip(tr("This sets the selected output device to the sample rate of the content being played"));
+#endif
+
 	QFormLayout *layout = new QFormLayout(this);
 	layout->addRow(enabledB);
 	layout->addRow(tr("Playback device") + ": ", devicesB);
 	layout->addRow(tr("Delay") + ": ", delayB);
+	layout->addRow(bitPerfect);
```
zaps166: Crash, please add `#ifdef Q_OS_MACOS`

src/modules/PortAudio/PortAudio.cpp
```diff
@@ -88,15 +88,23 @@ ModuleSettingsWidget::ModuleSettingsWidget(Module &module) :
...
 }
 
 void ModuleSettingsWidget::saveSettings()
 {
 	sets().set("WriterEnabled", enabledB->isChecked());
 	sets().set("OutputDevice", devicesB->currentIndex() == 0 ? QString() : devicesB->currentText());
 	sets().set("Delay", delayB->value());
+	sets().set("BitPerfect", bitPerfect->isChecked());
```
zaps166: Crash, please add `#ifdef Q_OS_MACOS`

src/modules/PortAudio/PortAudio.hpp
```diff
@@ -53,4 +53,5 @@ class ModuleSettingsWidget final : public Module::SettingsWidget
 	QCheckBox *enabledB;
 	QComboBox *devicesB;
 	QDoubleSpinBox *delayB;
+	QCheckBox *bitPerfect;
```
zaps166: Please add `#ifdef Q_OS_MACOS`

src/modules/PortAudio/PortAudioWriter.cpp
```diff
@@ -43,10 +46,21 @@ PortAudioWriter::PortAudioWriter(Module &module) :
 	outputParameters.sampleFormat = paFloat32;
 	outputParameters.hostApiSpecificStreamInfo = nullptr;
 
+#ifdef Q_OS_MACOS
+	coreAudioDevice = nullptr;
```
zaps166: This is already initialized in header file :)

src/modules/PortAudio/PortAudioWriter.cpp
```diff
@@ -43,10 +46,21 @@ PortAudioWriter::PortAudioWriter(Module &module) :
...
 	SetModule(module);
 }
 PortAudioWriter::~PortAudioWriter()
 {
+#ifdef Q_OS_MACOS
+	if (coreAudioDevice)
+	{
+		coreAudioDevice->ResetNominalSampleRate();
+	}
+	delete coreAudioDevice;
```
zaps166: `delete nullptr` is no-op, but if you have `if`, you can move this above :)

src/modules/PortAudio/PortAudioWriter.cpp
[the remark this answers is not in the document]
zaps166: I'll check this code at runtime :)

src/modules/PortAudio/PortAudioWriter.cpp
```diff
@@ -203,12 +240,36 @@ bool PortAudioWriter::open()
...
+	}
+#endif
 	PaStream *newStream = nullptr;
 	if (Pa_OpenStream(&newStream, nullptr, &outputParameters, sample_rate, 0, paDitherOff, nullptr, nullptr) == paNoError)
 	{
 		stream = newStream;
 		outputLatency = Pa_GetStreamInfo(stream)->outputLatency;
 		modParam("delay", outputLatency);
+#ifdef Q_OS_MACOS
+		modParam("rate", Pa_GetStreamInfo(stream)->sampleRate);
```
zaps166: This parameter is set by `AudioThr` and shouldn't be changed here. Maybe use class member variable if it is needed for `PortAudioWriter::processParams`? And shouldn't it be inside `if (bitPerfectMode)`?

src/modules/PortAudio/PortAudioWriter.cpp
```diff
@@ -203,12 +240,36 @@ bool PortAudioWriter::open()
...
+		if (bitPerfectMode)
+		{
+			const QString devName = QString::fromUtf8(Pa_GetDeviceInfo(outputParameters.device)->name);
```
zaps166: `QString::fromUtf8` does nothing, you can remove this.

src/modules/PortAudio/PortAudioWriter.cpp
```diff
@@ -203,12 +240,36 @@ bool PortAudioWriter::open()
...
+			AudioDeviceList::DeviceDict devDict = AudioDeviceList().GetDict();
```
zaps166: `const AudioDeviceList::DeviceDict`

src/modules/PortAudio/PortAudioWriter.cpp
```diff
@@ -203,12 +240,36 @@ bool PortAudioWriter::open()
...
+			if (devDict.contains(devName))
+			{
+				coreAudioDevice = AudioDevice::GetDevice(devDict[devName], false, coreAudioDevice);
+				if (coreAudioDevice && bitPerfectMode)
```
zaps166: `if (coreAudioDevice)`, `bitPerfectMode` is checked earlier

src/modules/PortAudio/PortAudioWriter.cpp
```diff
@@ -203,12 +240,36 @@ bool PortAudioWriter::open()
 
 bool PortAudioWriter::openStream()
 {
+#ifdef Q_OS_MACOS
+	bool bitPerfectMode = sets().getBool("BitPerfect");
+	if (bitPerfectMode && (outputDevice.isEmpty() || outputDevice == "Default"))
```
zaps166: IIRC default device is always empty/null. Moreover default device is translated on the list. So `outputDevice == "Default"` not needed.

src/modules/PortAudio/PortAudioWriter.cpp
```diff
@@ -203,12 +240,36 @@ bool PortAudioWriter::open()
...
+	{
+		int chn = getParam("chn").toInt();
+		outputParameters.device = PortAudioCommon::getDeviceIndexForOutput(outputDevice, chn);
```
zaps166: This is done in `PortAudioWriter::processParams`, soon before calling this function, IMHO not needed here.

src/modules/PortAudio/PortAudio.cpp
RJVB: oops...

src/modules/PortAudio/PortAudio.cpp
RJVB: oops, bis.

Shouldn't we in fact create the bitPerfect checkbox on all platforms but show it only on Mac? That would be the easiest way to ensure that the settings file preserves everything, even if you share it across different platforms.

src/modules/PortAudio/PortAudioWriter.cpp
RJVB: Well, you asked for it :p

On second thought, I'm moving the initialisation up so that all initialised member variables are handled in the same way in the same location. I don't really like initialising variables in a header; too easy to forget.

src/modules/PortAudio/PortAudioWriter.cpp
RJVB: Are you certain it is always being done and that the setting will never be changed? (Seems unlikely but can't shake the feeling I put the block there for a reason).

I'll re-test.

src/modules/PortAudio/PortAudioWriter.cpp
RJVB: Again, I'll have to re-test to see if I can remember why I added the statement.

src/modules/PortAudio/PortAudioWriter.cpp
RJVB: I'm used to KDE where the ctors taking C strings have been disabled...

src/modules/PortAudio/PortAudioWriter.cpp
zaps166: > Well, you asked for it

I asked for initialization in header, because QMPlay2 uses now C++11 and we can use new features. Initializing POD data in constructor to its default values (0) makes no sense, so hopefully C++11 helps us allowing to initialize in header :)

src/modules/PortAudio/PortAudio.cpp
zaps166: bitPerfect is available on Mac, so no need to store it on other platforms. If somebody adds it for Windows, then it will be changed.

RJVB:
>bitPerfect is available on Mac, so no need to store it on other platforms. If somebody adds it for Windows, then it will be changed.

How would you feel if you opened, say, a word processing document on a different computer that doesn't have any printers configured, and you lost all print settings for the document?

It's a bit of a far-fetched example, but it shows the principle that we adhere to in a number of other cross-platform projects I collaborate on. You don't know what use cases users might invent, so take care not to trash settings that don't concern you (IOW, don't trespass ;) ).
With QtCurve for instance that requires quite a bit of extra code, here it's maybe 1 extra instruction (to hide the widget) and probably less #ifdefs.

Just saying.

RJVB:
>I asked for initialization in header, because QMPlay2 uses now C++11 and we can use new features.

New features aren't always progress. Inline initialisation is perfectly fine if you have an in-file class definition, doing it in a different file is impractical IMHE (humble experience :)) and thus likely to lead to errors.

>Initializing POD data in constructor to its default values (0) makes no sense, so hopefully C++11 helps us allowing to initialize in header :)

In other words, it's not even necessary to initialise it, so why bother with it?

zaps166:
> How would you feel if you opened, say, a word processing document on a different computer that doesn't have any printers configured, and you lost all print settings for the document?

If you open e.g. MKV file in QMPlay2 you don't loose anything.

---

Who needs not working checkbox with a comment that it doesn't work 😄 And who will copy configuration files from Mac to e.g. Windows? Moreover not saving this doesn't mean that this setting will be removed from configuration file - it won't change its value; nothing is lost.

RJVB:
On Tuesday February 06 2018 12:12:47 Błażej Szczygieł wrote:

> Who needs not working checkbox with a comment that it doesn't work 

My proposition was to hide it, so the user wouldn't even know there's a checkbox.

> And who will copy configuration files from Mac to e.g. Windows?

I exchange settings files between Mac and Linux all the time. And don't you provide a portable install version? It'd be nice if you could have a single removable drive with content and the possibility to have Mac, MSWindows and Linux versions of the player that all share the same settings, no? Plug in the drive, launch the player and off you go, no matter what platform you're on.

> Moreover not saving this doesn't mean that this setting will be removed from configuration file - it won't change its value; nothing is lost.

If that's the case then my proposition is moot

zaps166:
> I exchange settings files between Mac and Linux all the time

Not saving this on Linux doesn't corrupt your config files. If bit-perfect is `true` on mac, it'll remain true on Linux.

src/modules/PortAudio/CMakeLists.txt
```diff
@@ -13,6 +13,14 @@ set(PortAudio_SRC
     PortAudioWriter.cpp
 )
 
+if(APPLE)
+    list(APPEND PortAudio_SRC
+        3rdparty/CoreAudio/AudioDevice.mm
+        3rdparty/CoreAudio/AudioDeviceList.cpp
+    )
+    include_directories(3rdparty/CoreAudio)
```
zaps166: Not needed, and include:
```c++
#include "3rdparty/CoreAudio/AudioDeviceList.h"
#include "3rdparty/CoreAudio/AudioDevice.h"
```

src/modules/PortAudio/PortAudio.cpp
zaps166: Missing `#ifdef Q_OS_MACOS` - it doesn't compile on Linux/Windows.

src/modules/PortAudio/PortAudioWriter.cpp
```diff
@@ -24,6 +24,9 @@
 #endif
 
 #ifdef Q_OS_MACOS
+	#include <QDebug>
```
zaps166: `QDebug` not needed in this PR

src/modules/PortAudio/3rdparty/CoreAudio/AudioDevice.h
```diff
@@ -0,0 +1,136 @@
...
+#ifndef __AudioDevice_h__
+#define __AudioDevice_h__
+
+#include <CoreServices/CoreServices.h>
+#include <CoreAudio/CoreAudio.h>
+
+#include <qglobal.h>
+
+#ifndef DEPRECATED_LISTENER_API
+#	if !QT_MACOS_PLATFORM_SDK_EQUAL_OR_ABOVE(__MAC_10_11)
```
zaps166: Does not compile:
```
/Users/tomo/Documents/QMPlay2/src/modules/PortAudio/3rdparty/CoreAudio/AudioDevice.h:58: error: token is not a valid binary operator in a preprocessor subexpression
#       if !QT_MACOS_PLATFORM_SDK_EQUAL_OR_ABOVE(__MAC_10_11)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
```

src/modules/PortAudio/3rdparty/CoreAudio/AudioDevice.mm
```diff
@@ -0,0 +1,641 @@
...
+                    // update the rate we should reset to
+                    dev->SetInitialNominalSampleRate(sampleRate);
+                    if (!dev->listenerSilentFor) {
+                        NSLog(@"%@\n\tkAudioDevicePropertyActualSampleRate=%g\n", msg, sampleRate);
+                    }
+                }
+                break;
+            default:
+                if ((dev && !dev->listenerSilentFor)) {
+                    NSLog(msg);
```
zaps166:
```
/Users/tomo/Documents/QMPlay2/src/modules/PortAudio/3rdparty/CoreAudio/AudioDevice.mm:156:27: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
                    NSLog(msg);
                          ^~~
/Users/tomo/Documents/QMPlay2/src/modules/PortAudio/3rdparty/CoreAudio/AudioDevice.mm:156:27: note: treat the string as an argument to avoid this
                    NSLog(msg);
                          ^
                          @"%@", 
1 warning generated.
```

src/modules/PortAudio/PortAudioWriter.cpp
```diff
@@ -33,6 +36,9 @@ PortAudioWriter::PortAudioWriter(Module &module) :
 	stream(nullptr),
 	sample_rate(0),
 	err(false)
+#ifdef Q_OS_MACOS
+	, coreAudioDevice(nullptr)
```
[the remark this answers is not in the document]
RJVB: Well, you're the boss here :)
I do notice though that none of the examples in that link use separate definition and declaration (header and implementation) files ;)

src/modules/PortAudio/3rdparty/CoreAudio/AudioDevice.h
RJVB: What Qt version? I'm not certain if compiling Qt apps with just the CLT is supported officially but if it is that should not make a difference. Those macros are pure Qt but the 10_11 ones were added in June 2015 only.

Does copying the definition for __MAC_10_11 and/or MAC_OS_X_VERSION_10_11 solve the issue (I'm including both but hope only __MAC_10_11 is required)?

src/modules/PortAudio/3rdparty/CoreAudio/AudioDevice.mm
RJVB: If you want to be bothered by that :)

src/modules/PortAudio/3rdparty/CoreAudio/AudioDevice.h
zaps166: `QT_MACOS_PLATFORM_SDK_EQUAL_OR_ABOVE` doesn't exist in Qt 5.6, but QMPlay2 must support Qt 5.6. `MAC_OS_X_VERSION_10_11` seems to not be used in project.

zaps166: You defined wrong macro (`QT_OSX_PLATFORM_SDK_EQUAL_OR_ABOVE` != `QT_MACOS_PLATFORM_SDK_EQUAL_OR_ABOVE`).

I think it is better to use `__MAC_OS_X_VERSION_MAX_ALLOWED` directly. Any why there is need to use `DEPRECATED_LISTENER_API`?

RJVB:
On Sunday February 11 2018 22:00:40 Błażej Szczygieł wrote:

>You defined wrong macro (`QT_OSX_PLATFORM_SDK_EQUAL_OR_ABOVE` != `QT_MACOS_PLATFORM_SDK_EQUAL_OR_ABOVE`).

From the 5.9 qglobal.h:
```
// Compatibility synonyms, do not use

#  define QT_OSX_PLATFORM_SDK_EQUAL_OR_ABOVE(osx) QT_MACOS_PLATFORM_SDK_EQUAL_OR_ABOVE(osx)
```

QT_OSX_PLATFORM_SDK_EQUAL_OR_ABOVE is the old form of QT_MACOS_PLATFORM_SDK_EQUAL_OR_ABOVE and will hopefully work with Qt 5.6 .
I did make a stupid error (again...) implementing the fallback; that should be fixed now.

>
>I think it is better to use `__MAC_OS_X_VERSION_MAX_ALLOWED` directly. Any why there is need to use `DEPRECATED_LISTENER_API`?

That will lead to similar code because Apple hasn't been defining the version macros several versions in advance. So maybe it's better because the rest of the class doesn't depend on Qt but that's about the only difference I think.


I'm not certain why it's needed to use this API on 10.9 but since you support Qt 5.6 QMPlay2 should run also on older OS X versions (down to 10.7 IIRC). The new API was introduced in 10.7 but that doesn't mean it worked perfectly. CoreAudio is notorious for its bad (rather, absent) documentation, so in practice you use what works. I notice that PortAudio also uses the old API.

(Also look at how long it took for AVFoundation to come even close to feature-completion AFTER QuickTime was deprecated.)

If you prefer we can just use the old API in the upstream implementation and keep both plus the switch in my fork for testing until I figure out how to get that new API to work on all supported OS versions.

BTW, you probably saw that the listener is only there to catch cases where the user changes the device sample rate via the OS while QMPlay2 is playing in bitperfect mode (or when another application does this).
I'm not sure I consider that a crucial feature, do you? We can also document that any 3rd party changes to the samplerate will be lost when playback stops and restores the initial samplerate.
After all you probably shouldn't make such changes during playback (except for testing, and then QMplay2 should probably just restore that initial samplerate)?

RJVB:
Yay :)

zaps166:
Remaining bugs, missing docs, etc, can be fixed later :)

---

I've noticed that this code changes sample rate even if it is equal (e.g. 44100 -> 44100) according to logs. Is it correct? I cannot test changing samplerate on VirtualBox (only 44100 Hz is available).

> you probably saw that the listener is only(...)

I haven't read the entire Obj-C code :)

---

I want to merge this by squashing it into a single commit, OK?

RJVB:
> Remaining bugs, missing docs, etc, can be fixed later :)

Sure.

> I've noticed that this code changes sample rate even if it is equal (e.g. 44100 -> 44100) according to logs. Is it correct?

No, AudioDevice::SetNominalSampleRate() is actually a noop when the current samplerate is already the requested rate. The log doesn't say it changes the rate ;) Of course that was mostly for debugging purposes, if you want I can move the NSLog call into the if block that does the actual work.


> > you probably saw that the listener is only(...)
> 
> I haven't read the entire Obj-C code :)

Fortunately there's more than just the listener code :)

R.

zaps166:
I merged it in console and Github doesn't know that this is merged, hmm...

zaps166:
Maybe I should remove commit and merge here.

zaps166:
I don't know if an issue number triggers a merge or something else, but now it is OK :)

RJVB:
That's what counts, no? :)

zaps166:
@RJVB Thanks for your PR :)

Issue #132, opened 2018-01-01
Policy CMP0071 related warning with cmake-3.10.1


reagentoo:
```
-- Configuring done
CMake Warning (dev) in src/gui/CMakeLists.txt:
  Policy CMP0071 is not set: Let AUTOMOC and AUTOUIC process GENERATED files.
  Run "cmake --help-policy CMP0071" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  For compatibility, CMake is excluding the GENERATED source file(s):

    "/var/tmp/portage/media-video/qmplay2-17.12.31/work/qmplay2-17.12.31_build/src/gui/ui_SettingsGeneral.h"
    "/var/tmp/portage/media-video/qmplay2-17.12.31/work/qmplay2-17.12.31_build/src/gui/ui_SettingsPlayback.h"
    "/var/tmp/portage/media-video/qmplay2-17.12.31/work/qmplay2-17.12.31_build/src/gui/ui_SettingsPlaybackModulesList.h"
    "/var/tmp/portage/media-video/qmplay2-17.12.31/work/qmplay2-17.12.31_build/src/gui/ui_OSDSettings.h"

  from processing by AUTOMOC.  If any of the files should be processed, set
  CMP0071 to NEW.  If any of the files should not be processed, explicitly
  exclude them by setting the source file property SKIP_AUTOMOC:

    set_property(SOURCE file.h PROPERTY SKIP_AUTOMOC ON)

This warning is for project developers.  Use -Wno-dev to suppress it.
```
I suggest to insert
```
set_property(SOURCE ${GUI_FORM_HDR} PROPERTY SKIP_AUTOMOC ON)
```
after
https://github.com/zaps166/QMPlay2/blob/master/src/gui/CMakeLists.txt#L99

zaps166:
You can open a PR for this :)

PR #133, opened 2018-01-02
CMake: Append SKIP_AUTOMOC property for autogenerated headers

reagentoo:
For supress policy CMP0071 related warnings with >=cmake-3.10.1.
Issue: https://github.com/zaps166/QMPlay2/issues/132


zaps166:
Thanks!

PR #208, opened 2018-11-20
Updated Hungarian translation

notramo:


notramo:
@zaps166 Could you create a new release?
Void Linux only ships tagged releases.

zaps166:
@notramo Done (as a pre-release)

Issue #231, opened 2019-06-13
YouTube playback and download failing on more and more videos


RJVB:
I have always used QMPlay2 as my favourite YouTube browser; it is so much lighter and faster than going through the browser. Sadly I'm getting playback and downloading failures on more and more YouTube videos, and this has increased steeply recently (last few weeks).

It seems YT now serves "manifest" URLs like

```
https://manifest.googlevideo.com/api/manifest/dash/expire/1560441347/ei/ox0CXfTAKc7h1gae14vACg/ip/86.208.130.197/id/d953d116c709daa9/source/youtube/requiressl/yes/playback_host/r5---sn-25ge7nsd.googlevideo.com/mm/31%2C26/mn/sn-25ge7nsd%2Csn-4g5ednse/ms/au%2Conr/mv/m/pl/16/hfr/all/as/fmp4_audio_clear%2Cwebm_audio_clear%2Cwebm2_audio_clear%2Cfmp4_sd_hd_clear%2Cwebm2_sd_hd_clear/initcwndbps/808750/mt/1560419662/fvip/5/itag/0/sparams/expire%2Cei%2Cip%2Cid%2Csource%2Crequiressl%2Chfr%2Cas%2Citag/sig/ALgxI2wwRQIhALt9zxWST-W2TESIEII0c8Z3bvIfqpg6lRMw-RDIT5pPAiBE5hRbgQWoqV0B125_xCWmOjWGiCmatUPa5appp8regQ%3D%3D/lsparams/playback_host%2Cmm%2Cmn%2Cms%2Cmv%2Cpl%2Cinitcwndbps/lsig/AHylml4wRQIhAPSzZqJ-Mz30gcyX5lBjLsMOF3-tZHI2EKmqyEt5jRy0AiADcFKkVGwFkSyCNWr3Y2G3Gx-w1wXPbPnaIXtZRix-AA%3D%3D
```

These appear to be valid only for a limited duration, too, so I am attaching the contents
[...]

PR #232, opened 2019-06-15
Quick fix for the current YouTube playback/download problem.

RJVB:
This modifies the `youtube-dl -g -j` parsing so a list of actual
content streams is obtained, sorted to favour the higher qualities
(read: larger vertical resolution).

It is a quick fix because it will not provide the user with a collapsable
list of resolutions under the video thumbnail. Instead the video will
be treated as content with multi streams (and downloaded to a mkv
container); the user can select video and/or audio streams at will via
the Information widget. Not the most convenient (user preferences for
resolution are ignored) but infinitely better than getting an error on
trying to open an URL that is not a media stream.
The resulting address (as would be reported in the Information widget)
is not one you can do anything with, so the YouTube page address is
reported instead.

TODO: exclude stream URLs that cannot be played anyway and which will
presumably increase processing overhead before the streams starts
playing.

Fixes: #231

### DIFF
--- a/src/qmplay2/YouTubeDL.cpp
+++ b/src/qmplay2/YouTubeDL.cpp
@@ -25,6 +25,7 @@
 #ifdef Q_OS_WIN
     #include <Functions.hpp>
 #endif
+#include <Module.hpp>
 
 #include <QReadWriteLock>
 #include <QJsonDocument>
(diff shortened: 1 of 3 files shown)


src/qmplay2/YouTubeDL.cpp
```diff
@@ -308,26 +315,38 @@ QStringList YouTubeDL::exec(const QString &url, const QStringList &args, QString
...
-                result.removeAt(i);
             }
         }
 
+        // construct the return list
+        QStringList retList(result.at(0));
+        auto urlList = height2Url.values().toStdList();
+        // reverse-sort the URL list so that the higher resolutions get priority
+        urlList.reverse();
+        for (const auto streamUrl : urlList)
```
zaps166: Maybe you can iterate over `height2Url` and prepend values instead of copying them to `QList<QString>` and then conversion to std linked list?

src/qmplay2/YouTubeDL.cpp
```diff
@@ -308,26 +315,38 @@ QStringList YouTubeDL::exec(const QString &url, const QStringList &args, QString
...
+        {
+            retList << streamUrl;
+        }
+
+        // clear the status bar message, we're done.
+        emit QMPlay2Core.statusBarMessage("", 0);
```
zaps166: You can use null `QString()`

RJVB:
>Maybe you can iterate over `height2Url` and prepend values instead of copying them to `QList<QString>` and then conversion to std linked list?

Yeah, that works too. I was so focussed on finding an atomic reverse-list operation that I didn't realise I could just as well roll it into the loop I had to use anyway to create the return list ...

Don't you think that it'd be nicer to display the actual youtube page address in the Information widget, instead of the huge blurb which you cannot do anything useful with?

I found the place where the info content is constructed but I cannot figure out if and how the page address is available there?

RJVB:
Maybe it would be an easy adaptation to exclude streams that correspond to a resolution that the user has not configured (= has elected to ignore), if that information is available?

I just came across a video that has these streams

```
> youtube-dl -F "https://www.youtube.com/watch?v=KT7j-pYUJ5I"
[youtube] KT7j-pYUJ5I: Downloading webpage
[youtube] KT7j-pYUJ5I: Downloading video info webpage
[info] Available formats for KT7j-pYUJ5I:
format code  extension  resolution note
249          webm       audio only DASH audio   61k , opus @ 50k, 3.59MiB
250          webm       audio only DASH audio   76k , opus @ 70k, 4.41MiB
171          webm       audio only DASH audio  118k , vorbis@128k, 6.63MiB
140          m4a        audio only DASH audio  128k , m4a_dash container, mp4a.40.2@128k, 8.70MiB
251          webm       audio only DASH audio  134k , opus @160k, 7.93MiB
278          webm       256x144    144p   97k , webm container, vp9, 25fps, video only, 6.12MiB
160          mp4        256x144    144p  113k , avc1.4d400c, 25fps, video only, 6.37MiB
242          webm       426x240    240p  223k , vp9, 25fps, video only, 12.30MiB
133          mp4        426x240    240p  251k , avc1.4d4015, 25fps, video only, 14.00MiB
243          webm       640x360    360p  409k , vp9, 25fps, video only, 23.10MiB
134          mp4        640x360    360p  637k , avc1.4d401e, 25fps, video only, 28.88MiB
244          webm       854x480    480p  775k , vp9, 25fps, video only, 42.61MiB
135          mp4        854x480    480p 1167k , avc1.4d401e, 25fps, video only, 59.37MiB
247          webm       1280x720   720p 1518k , vp9, 25fps, video only, 91.75MiB
136          mp4        1280x720   720p 2343k , avc1.4d401f, 25fps, video only, 118.90MiB
248          webm       1920x1080  1080p 2702k , vp9, 25fps, video only, 174.19MiB
137          mp4        1920x1080  1080p 4412k , avc1.640028, 25fps, video only, 241.31MiB
271          webm       2560x1440  1440p 8896k , vp9, 25fps, video only, 581.13MiB
313          webm       3840x2160  2160p 17873k , vp9, 25fps, video only, 1.17GiB
18           mp4        640x360    medium , avc1.42001E, mp4a.40.2@ 96k, 45.72MiB
43           webm       640x360    medium , vp8.0, vorbis@128k, 56.51MiB
22           mp4        1280x720   hd720 , avc1.64001F, mp4a.40.2@192k (best)
```

needless to say that not every computer will like 3840x2160 resolutions...

zaps166:
Could you add example URLs which doesn't work without this patch?

RJVB:
In the commit message? Recently most videos didn't work anymore, that's why I filed the issue and made this patch.
I can of course add examples but as you know they can disappear at any time (or even start working again if YouTube changes again).

There's one in the linked issue (#231) btw.

zaps166:
I have some problems recently with YouTube on QMPlay2. On some videos FFmpeg buffers all DASH streams. On some other videos I don't have collapsible tree and I don't have subtitles.

With this patch I can't play videos which I can play without this.

The video https://www.youtube.com/watch?v=LXb3EKWsInQ plays with and w/o the patch, but with this patch FFmpeg buffers all stream with different A/V codecs and video sizes. Without there is single 4k stream.

The video plays before this patch, but now doesn't play: https://www.youtube.com/watch?v=Y_TTsEfFrHA

zaps166:
I can try to fix it, but the major problem is `youtube-dl` - it is a Python script (slow). Moreover it must be decompressed and compiled into a Python bytecode every time before it start. On my i7-8700k CPU it starts ~870ms, so I can't use `youtube-dl` everywhere :disappointed: And I can't find how to run it once and keep it in background and send commands e.g. via `stdin`.

RJVB:
I’ll look into that then, I must admit I have only tested my change on videos that wouldn’t play. It shouldn’t be hard to test for the “legacy” json format to decide which code path to use, so that most videos will play.

RJVB:
Re python script: if that has been written with a main function that calls various working functions then you should be able to do the compilation once and then invoke the bytecode. I know what difference this can make if you call simple functions very often, am a bit surprised that this script is so expensive to compile.

RJVB:
> The video https://www.youtube.com/watch?v=LXb3EKWsInQ plays with and w/o the patch, but with this patch FFmpeg buffers all stream with different A/V codecs and video sizes. Without there is single 4k stream.

True, but with my patch it becomes possible at least to select a lower res stream if your system (or connection) cannot handle those crazy resolutions.

I'll post an updated patch tomorrow.

RJVB:
I pushed a fix that re-enables the legacy code when exec() is called either with a list of tags (-j option; AFAICT this only happens for videos that can actually be parsed by the old code, presumably because the YT extension does its own parsing and got somewhere with that). I've also added a filter that should reduce the required bandwidth by selecting only the streams that correspond to user-selected resolutions.

Please have a look if I'm getting those preferences in the most efficient way (apart from the potential caching of the extension module pointer as a static member variable).

> Moreover it must be decompressed and compiled into a Python bytecode every time before it start.

Isn't there a way to avoid that overhead?
- decompress, byte-compile, recompress
- decompress and invoke the entrypoint (or write one yourself)?
Sadly I've never come across this kind of packaging before and I can't figure out how to use the code after decompressing it...

I agree the script is expensive (on a lowly N3150), mostly due to the required decompression
```
> time repeat 5 ~/.qmplay2/youtube-dl --help > & /dev/null  
19.204 user_cpu 1.964 kernel_cpu 0:19.71 total_time 107.3%CPU {39920M 0F 217606R 1576835I 0O 267w 5350c}
```

RJVB:
Well, I just missed the wrapper script from the zipped bundle. Unzipped the thing is a lot (4x) faster:

```
> time repeat 5 python /tmp/__main__.py --help > & /dev/null
4.600 user_cpu 1.097 kernel_cpu 0:05.25 total_time 108.3%CPU {35140M 0F 63524R 61190I 0O 252w 1419c}
```

This will probably be noticeable, on slower hosts at least.

Do you already have builtin support for unpacking a zip archive? If so it should be trivial to unpack the compressed script after each successful download and then call it as above.

zaps166:
> Do you already have builtin support for unpacking a zip archive? If so it should be trivial to unpack the compressed script after each successful download and then call it as above.

I can use `unzip` command if exists, but updates will be more difficult. And what about Windows? On Window I need to embed Python and unzip in my installer :smile: Maybe I'll try with unzip, but not for upcoming release.

zaps166:
`YouTube` and `YouTubeDL` classes has changed, so this PR is no longer valid.
Anyway, thanks for the quick fix.

PR #235, opened 2019-07-24
Remove unused preprocessor directives in MediaBrowser.cpp

Martchus:
The unused libavutil header breaks the build in case libavutil headers are not installed in a standard location.

By the way, I'm currently trying to build this under Tumbleweed because the official package crashes on startup. Maybe the recently enabled LTO causes problems.

### DIFF
--- a/src/modules/Extensions/MediaBrowser.cpp
+++ b/src/modules/Extensions/MediaBrowser.cpp
@@ -50,10 +50,6 @@
 
 #include <algorithm>
 
-#include <libavutil/cpu.h>
-#ifdef QMPLAY2_CPU_X86_32
-#endif
-
 Q_LOGGING_CATEGORY(mb, "MediaBrowser")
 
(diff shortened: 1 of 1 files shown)


zaps166:
Thanks.

Yes, LTO can cause a crash. IMHO LTO is good if you link the entire application into a single executable. It is also good option for microcontrollers.

Martchus:
So this is a known issue? I'm wondering why it would crash.

I'm going to disable LTO in the openSUSE package then for now. Note that it can also make sense for a Linux distribution to enable LTO. It at least reduced the total binary size (see openSUSE-factory mailing list).

zaps166:
I also use LTO for `qmplay2` AUR package. I know a single case that somebody has a crash with LTO.

Yes, LTO also reduces binary size, but I know that it can cause problems with RTTI and maybe with exceptions when using with executables and `*.so` objects (e.g. `std::any` might not work properly between exec and lib). QMPlay2 doesn't use both of them, so it should be OK.

Martchus:
Potential problems with RTTI sound quite bad. By the way, thanks for writing this player. It is the best media player for the desktop I've seen so far!

PR #262, opened 2019-11-03
Update Ukrainian translation file

Lan72:


zaps166:
Thanks for translation!

I updated translations once again, now it should be all for upcoming release :)

PR #291, opened 2020-02-26
VAAPIOpenGL: fix possible missing glEGLImageTargetTexture2DOES

arthurzam:
Hello.
I'm using QMPlay2 with `VA-API` decoder, and after 763b4806baad7e356fc62f3877919894b92f9b35 it started to crash with following error message:
`QMPlay2: symbol lookup error: /usr/lib64/qmplay2/modules/libFFmpeg.so: undefined symbol: glEGLImageTargetTexture2DOES`

After searching the code for this function name, I found it used only in `modules/FFmpeg/VAAPIOpenGL.cpp`.

This small change fixed for me (and it looks more consistent with the `m_egl` everywhere in this file).


zaps166:
Thanks!

PR #293, opened 2020-03-10
cmake: fix TagLib when first configure with USE_TAGLIB=ON

arthurzam:
When first configuring QMPlay2 with `USE_TAGLIB=ON`, it doesn't finds
or loads the paths, thus without including taglib libs.


zaps166:
Thanks, I missed it, but I think it should be `REQUIRED` if enabled :)

PR #306, opened 2020-04-03
Renamed Strings in Playlist

zo-shin:
1 - Renamed some confusing strings in the playlist.
2 - Fixed a typo.


zaps166:
Thanks. I'll think about it, but not for the upcoming release.

zo-shin:
Commit updated.

PR #307, opened 2020-04-12
Update installation instructions for openSUSE, add installation instructions for Fedora using RPMFusion

DarkWav:
So, I recently had some time so I thought it would be nice to polish the install instructions for openSUSE and update them for 15.2. I also added some instructions for Fedora 32, 31 and rawhide which use RPMFusion and include the installation of any necessary multimedia codecs.
I hope you like the new layout, any improvement suggestions are appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,11 @@ Table of Contents
 - run the following command: `yay -S qmplay2`
 
 #### Easy installation on openSUSE
+- Don't mix FFmpeg from different repositories!
 ##### For openSUSE Leap 15.2:
-- Run the following commands:
+<details>
+<summary>Run the following commands:</summary>
+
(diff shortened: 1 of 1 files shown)


zaps166:
Could you replace HTML tags with markdown syntax (especially `<pre>`)? It might not display properly outside GitHub. Thanks!

DarkWav:
I will try my best, thanks for the suggestion!

DarkWav:
OK, I switched the instructions to standard markdown. Sadly no spoilers there, but I managed to unify the Fedora instructions and removed Leap 15.1 as it will be EOL fairly soon.

zaps166:
Thanks!

PR #330, opened 2020-05-08
Added New Strings to zh.ts

zo-shin:
Added four strings of the drop-down menu of radio search to zh.ts


zaps166:
This won't work, I've made code changes for translations.

zo-shin:
Commit updated.

PR #350, opened 2020-08-07
enable support for sdp file

fay2003hiend:


fay2003hiend:
Hi

This could fix the error
```
Protocol 'rtp' not on whitelist 'file,crypto,data'!
```
when openning a sdp file like below
```
c=IN IP4 127.0.0.1
m=video 6000 RTP/AVP 96 
a=rtpmap:96 H264/90000
```

src/modules/FFmpeg/FormatContext.cpp
```diff
@@ -719,6 +719,13 @@ bool FormatContext::open(const QString &_url, const QString &param)
     // Useful, e.g. CUVID decoder needs valid PTS
     formatCtx->flags |= AVFMT_FLAG_GENPTS;
 
+    // enable sdp file & reduce udp latency
+    if (url.endsWith("sdp"))
+    {
+        av_dict_set(&options, "protocol_whitelist", "file,udp,rtp", 0);
```
zaps166, line 725: Is it possible to get whitelist and append rtp?

fay2003hiend: I think it is safe to do so, if the `formatCtx` is only temporary to this playback session

fay2003hiend: I just did a quick check, `formatCtx->codec_whitelist` is NULL before we call `formatCtx = openThr->getFormatCtx();`
so it seems no way to fetch and append the white list.

inside Ffmpeg source code avio.c:195 it seems to have a hard coded default_whitelist which is "file,crypto,data" for files if user supplies NULL

probably I can change it to "file,crypto,data,udp,rtp" in my commit?

fay2003hiend:
I've also added some options to allow lower latency of Ffmpeg backend, that helps quite a lot for live streams.

src/modules/FFmpeg/FormatContext.cpp
zaps166: > file,crypto,data,udp,rtp

ok

src/modules/FFmpeg/FFDec.cpp
```diff
@@ -81,6 +81,7 @@ AVCodec *FFDec::init(StreamInfo &streamInfo)
 }
 bool FFDec::openCodec(AVCodec *codec)
 {
+    codec_ctx->flags |= AV_CODEC_FLAG_LOW_DELAY;
```
zaps166, line 84: I have to check it... Maybe apply only for live streams? Are there any side effects (bugs, higher CPU/GPU usage, poor multithreading, etc)? What about local files?

fay2003hiend: sorry I could not really tell if that will cause any trouble. I modified a few more files to introduce `bool is_local` in StreamInfo, so that FFDec can recognise and enable LOW_DELAY flag for only live streams, is that what you prefer?

zaps166:
I'll check this soon, thanks!

zaps166:
Please check if 61ac4188869faf348184adb3ea3a2f01d2919bd2 is ok.

fay2003hiend:
> Please check if [61ac418](https://github.com/zaps166/QMPlay2/commit/61ac4188869faf348184adb3ea3a2f01d2919bd2) is ok.

thanks for the update, I'll verify it soon

PR #366, opened 2020-11-07
add japanese language file

harujisaku:
add japanese language file.

I'm not good at English, so I'm sorry if it is difficult for you to understand my English.
and It may be wrong because it is the first pul request.


zaps166:
Thanks!

PR #370, opened 2020-12-08
Fixed Language String Error

zo-shin:
Fixed a string error in Japanese


harujisaku:
oops! thank you

zo-shin:
You're welcome.

PR #386, opened 2020-12-29
Strings Amendments

zo-shin:
Renamed some confusing strings.

### DIFF
--- a/lang/de.ts
+++ b/lang/de.ts
@@ -580,8 +580,8 @@
         <translation>Download stoppen</translation>
     </message>
     <message>
-        <source>Size</source>
-        <translation>Größe</translation>
+        <source>File size</source>
+        <translation>Dateigröße</translation>
     </message>
     <message>
         <source>Download complete</source>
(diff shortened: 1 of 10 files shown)


src/gui/DemuxerThr.cpp
```diff
@@ -907,7 +907,7 @@ void DemuxerThr::emitInfo()
                     videoStreams += "<li><b>" + tr("title") + ":</b> " + streamInfo->title + "</li>";
                 if (!streamInfo->codec_name.isEmpty())
                     videoStreams += "<li><b>" + tr("codec") + ":</b> " + streamInfo->codec_name + "</li>";
-                videoStreams += "<li><b>" + tr("size") + ":</b> " + QString::number(streamInfo->width) + "x" + QString::number(streamInfo->height) + "</li>";
+                videoStreams += "<li><b>" + tr("resolution") + ":</b> " + QString::number(streamInfo->width) + "x" + QString::number(streamInfo->height) + "</li>";
```
zaps166, line 910: It is a video frame size, so it is a good text.

zo-shin: “Size” has more than one meanings. I think “resolution” is more straightforward than it. If you insist, I can revert this string.

PR #388, opened 2021-01-07
pitch

Turacana:
As the video speed increases, the audio quality deteriorates. The voices sound like people have swallowed a liter of helium. Please do it better.
For example, like on YouTube.


zaps166:
What does this PR do exactly? I don't want to merge `qt5.8` branch into `master`.

PR #439, opened 2021-08-28
Add initial support for Visual Studio

DecomoZ:
This patch adds a compilation support for Visual Studio.
I made and tested it on Visual Studio 2019 with vcpkg and made sure VS was successful in building and debugging QMPlay2.
It can only achieve minimal support at the moment, but I intend to carry on working.

Changes:
- Add CMakeLists.txt settings for Visual Studio.
- Change some codes to avoid compilation error in Visual Studio.
- Add build information by Visual Studio to README.md.


zaps166:
Nice work, but isn't better to add clang-cl only support (which uses MSVC headers/libs) instead of Visual Studio compiler which is less compatible and produces worse machine code?

What about inline assembly in yadif deinterlacing filter? Is it possible to compile using cl and/or clang-cl? If not - this filter doesn't make any sense on CPU (C++ code is many times slower).

PR #461, opened 2021-11-19
Update zh_TW.ts

lks-wu:
The work of zh_TW translation started from the very first line of source English, long hours of editing time until starting a pull request.

Earlier this week, I wanted to make one small revision but was shocked by 600+ lines of changes in zh_TW.ts ...
Therefore I feel that I need to clarify something.

In Taiwan, we have lots of language usage standards and habits different from China.
For instance, in zh_CN.ts and my version of zh_TW.ts
           the word "abort"     translation in zh_CN is 「終止」 and in zh_TW is 「中止」,
meanwhile, the word "terminate"           zh_CN is 「終止」 and in  zh_TW is 「終止」.
And another one     "prefer"                            「首選」       v.s.          「偏好」.
These words are examples in terms of translation differences.
There are variants simply more than characters being simplified.
"Help" have always translated as 「說明」in application menu across all Taiwan Traditional Chinese (臺灣正體中文) localization, like vlc, smplayer, dolphin file manager... the list goes on.

Furthermore, there are too many differences in both educational/technical standards and daily usage habits.
While I was doing the translation work, I looked it up in the dictionary and picked the best translation from the website of
National Academy for Educational Research.
Website here: https://terms.naer.edu.tw/

I found some translations being replaced to ones in China, which are not our standard.
And I also spotted Simplified Chinese characters inserted in between which is not okay.

Concerned effort of help is welcome but better not this way.
Suggestion comments are better than force-pushed override.
I had to spend extra hours looking into what were revised and deciding whether to accept each one or not. Which was kind of and very tiresome.
Noob here, but zh_CN authors from China should not co-author zh_TW language unless familiar with and respect our differences.

After all above explanatory text, I would still praise the corrections of spacing,
I made some spacing choices I found out not in concord with common translation tradition...
like the spacing "關於QMPlay2" into "關於 QMPlay2"
and...
I  agree with those changes along with some others. So thank you... anyway...

### DIFF
--- a/lang/zh_TW.ts
+++ b/lang/zh_TW.ts
@@ -74,19 +74,19 @@
     <name>Add</name>
     <message>
         <source>&amp;Address</source>
-        <translation>位址</translation>
+        <translation>&amp;位址</translation>
     </message>
     <message>
         <source>&amp;Files</source>
-        <translation>檔案</translation>
+        <translation>&amp;檔案</translation>
     </message>
     <message>
         <source>&amp;Directory</source>
-        <translation>目錄</translation>
+        <translation>&amp;目錄</translation>
     </message>
     <message>
         <source>&amp;Add</source>
-        <translation>加入</translation>
+        <translation>&amp;加入</translation>
     </message>
 </context>
 <context>
@@ -206,7 +206,7 @@
     </message>
     <message>
         <source>Wallpaper in the main window</source>
-        <translation>主視窗壁紙</translation>
+        <translation>主視窗桌布</translation>
     </message>
     <message>
         <source>Transparency</source>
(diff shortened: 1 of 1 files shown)


zo-shin:
There are errors in your translation. Zoom in means 放大, zoom out means 縮小. You should keep them as what they were.

zo-shin:
And for the underlines in the menus, they are keyboard shortcut, which are not suitable for Chinese characters. You should not add &amp; to the translation. If it's a huge task for you, just ignore it. I'll make the changes later.

zaps166:
Can I merge?

zo-shin:
It can be merged. I'll correct it later .

PR #472, opened 2021-12-15
Changed Left Button to Primary Button

zo-shin:
The left button is not necessarily the primary one. "Primary" is more appropriate.


zaps166:
Can you use `obsolete` flag and keep old translations for other languages?

zaps166:
I didn't know it will not translate with `obsolete` type. Removed.

PR #482, opened 2022-01-25
show the YouTube video description in the Info panel

### LINKED ISSUE #465: [feature request] show description of YouTube videos
I just realised that the json document you get with `yt-dl -j <url>` also contains the video `description`, i.e. the text YouTube shows under the player.

It would be great if QMPlay2 could display that info (I find that much more interesting than, say, the full content URL ;)). As it is, I often have to open the YT page because I want to read the description, which is a waste of time that could be avoided.

It could be shown in the video's properties window (in copy'able form) and/or in the Information panel. In that panel, it could be abbreviated in a similar way as YouTube does because the panel will typically not be wide/big enough, and shown in complete form in a popup.

RJVB:
This is a proof-of-concept implementation for showing the description of YouTube videos in QMPlay2's Info panel.

The information is communicated via the YouTubeDL class instance, which required NOT doing the reset of the local instance in `YouTube::convertAddress()` (otherwise, `DemuxerThr::addSubtitleStream()` will see a NULL instance, hence the double check). There appears to be no side-effect of doing that (even with the reset, the instance is deleted only when the application exits).

A more proper implementation would probably use a dedicated class for storing metadata, including name, title etc. - rather than passing those around via function arguments. That however requires a much larger overhaul than I feel confident to handle ATM.

Fixes #465

### DIFF
--- a/src/gui/DemuxerThr.cpp
+++ b/src/gui/DemuxerThr.cpp
@@ -28,6 +28,7 @@
 #include <Demuxer.hpp>
 #include <Decoder.hpp>
 #include <Reader.hpp>
+#include <YouTubeDL.hpp>
 
 #include <QCryptographicHash>
 #include <QCoreApplication>
@@ -792,6 +793,7 @@ void DemuxerThr::addSubtitleStream(bool currentPlaying, QString &subtitlesStream
         streamMenuText.push_back(" - " + title);
     addStreamsMenuString(streamsMenu, streamCountStr, streamLink, currentPlaying, streamMenuText);
 }
+
 void DemuxerThr::emitInfo()
 {
     QString info;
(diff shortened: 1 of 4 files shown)


zaps166:
Thanks, I'll look into this code soon!

zaps166:
I see it's working, but for me it's too much text for information panel...
Best idea is to add it to Lyrics panel and add API to send text there. Do you want to try? Otherwise I can do it.

---

Please put brackets `{` in new line :smile:

RJVB:
>I see it's working, but for me it's too much text for information panel...

It can be a lot, indeed. And you'd probably want to wrap the text to the view width, something that I presume is already implemented for the lyrics tab.

>Best idea is to add it to Lyrics panel and add API to send text there. Do you want to try? Otherwise I can do it.

I could, but I think it's best if you take care of adding the API!

zaps166:
Done!

RJVB:
Thanks, I'll look ASAP!

PR #494, opened 2022-04-05
Improved Shortcut Scheme

### LINKED ISSUE #492: Improve Shortcut Scheme
Currently the menu shortcuts only support languages in Latin letters. My plan is to make all the shortcuts separated by moving them to brackets, like changing `&amp;Pause` to `Pause(&amp;P)`. In this case, all languages will be supported. Before the pull request is made, I'm wondering if anyone has better idea about that.

zo-shin:
Made all the shortcuts separated by moving them to brackets and other minor changes.
close #492

### DIFF
--- a/src/gui/MainWidget.cpp
+++ b/src/gui/MainWidget.cpp
@@ -348,7 +348,7 @@ MainWidget::MainWidget(QList<QPair<QString, QString>> &arguments)
 #if !defined Q_OS_MACOS && !defined Q_OS_ANDROID
     const bool menuHidden = settings.getBool("MainWidget/MenuHidden", false);
     menuBar->setVisible(!menuHidden);
-    hideMenuAct = new QAction(tr("&Hide menu bar"), menuBar);
+    hideMenuAct = new QAction(tr("Hide menu bar(&H)"), menuBar);
     hideMenuAct->setCheckable(true);
     hideMenuAct->setAutoRepeat(false);
     hideMenuAct->setChecked(menuHidden);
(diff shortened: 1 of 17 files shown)


src/gui/MainWidget.cpp
```diff
@@ -1,4 +1,4 @@
-/*
+﻿/*
```
zaps166, line 1: Please don't add UTF-8 BOM.

zaps166 (changes requested):

zaps166:
ehh doesn't compile

zaps166:
I fixed XML files, but it looks very bad:

Screenshot_20220425_004045

Reverted.

zo-shin:
What do you think is the problem? Is it OK if I just make changes to the language TS files only?

zaps166:
Changes to TS files are not a problem. I uploaded a screenshot of menu bar - it looks vary bad with brackets at the end: `Help(H)`, etc... Same with menu items.

zo-shin:
If that looks bad, do you have any better solution about the issue? It seems a common way to deal with shortcuts.

zaps166:
Maybe add it only for Chinese translations?

zo-shin:
All the languages should be solved in the same way, not only Chinese.

zaps166:
> All the languages should be solved in the same way, not only Chinese.

Chinese, Japanese. Example:


- QtCreator Chinese
Screenshot_20220430_123356

- QtCreator English
Screenshot_20220430_123427

Issue #496, opened 2022-04-18
Playback of mono recording in left channel only


RJVB:
A strange issue: QMPlay2 is one of a small handful of applications that have an issue with playback of mono audio over my Behringer UMC204HD. When that device is connected via one of the USB3 ports on my Belkin Thunderbolt dock, mono content plays back over the left channel only (from the affected applications).

QMPlay2 is evidently not at fault here, but there's clearly something it could be doing differently; neither VLC nor Apple's own QT Player are affected (but the system volume control feedback sound *is* affected).

zaps166:
I guess the problem is with mono playback on your device, it is playing mono over left channel (QMPlay2 and volume notification), but QuickTime and VLC probably mixes audio to stereo before playback.

I remember on Linux/Pulseaudio and 5.1 audio setup, mono was playing in center speaker only.

Try this setting: ![Screenshot_20220420_010046](https://user-images.githubusercontent.com/6239090/164114961-685d4339-a941-4913-a7b4-3fb808f8da03.png)

RJVB:
On Tuesday April 19 2022 16:01:15 Błażej Szczygieł wrote:

>I guess the problem is with mono playback on your device, it is playing mono over left channel (QMPlay2 and volume notification), but QuickTime and VLC probably mixes audio to stereo before playback.

Most applications apparently do. The curious thing is that it doesn't happen when I connect the device to one of the USB2 ports in/on my Mac. I am also seeing assertion warnings coming from a system library, so something is definitely off somewhere. It's odd enough that it caught the interest of a Behringer support engineer ;)

>I remember on Linux/Pulseaudio and 5.1 audio setup, mono was playing in center speaker only.

That actually makes sense, even if it's not what you expected! The centre speaker is supposed to be for "voice content", and it's in fact a recorded voice message that made me discover the issue.

>Try this setting: ![Screenshot_20220420_010046](https://user-images.githubusercontent.com/6239090/164114961-685d4339-a941-4913-a7b4-3fb808f8da03.png)

Ah, that looks like exactly the kind of workaround I was going to suggest (and looked for, in vain). Except I was thinking of an up-sampling approach, not one that converts everything.

I actually looked around in the code but realised I don't know much about how the actual sound data has to be sent. I did see something that looks like support for filters, but there is no evidence of audio filter support in the GUI? A fun solution to this mono problem would be to apply an (optional) filter that adds a pseudo-stereo effect ;)

RJVB:
[...]

PR #498, opened 2022-04-21
Make it possible to increase but not decrease the effective # of channels

RJVB:
Converts the ForceChannels checkbox from on/off to tri-state, where
partially checked means content with more than the selected number of
channels will not be altered; this allows for instance to promote mono
content to stereo without degrading quadraphonic or 5.1 content.
Previous ForceChannels==true settings are converted to fully checked.

Fixes: #496


zaps166:
Good idea!

RJVB:
>Good idea!

Thanks :)

PR #503, opened 2022-05-05
Reorganized Menu Shortcut

zo-shin:
1 - For languages in Latin letters, made shortcuts closest to English.
2 - For other languages, added shortcuts in brackets at the end.

### DIFF
--- a/lang/pt_BR.ts
+++ b/lang/pt_BR.ts
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.1" language="pt_BR">
+<TS version="2.1" language="pt_BR" sourcelanguage="en">
 <context>
     <name>ALSAWriter</name>
     <message>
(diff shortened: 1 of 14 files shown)


lang/pt_BR.ts
```diff
@@ -273,7 +273,6 @@
     </message>
     <message>
         <source>D&amp;epends on size</source>
-        <translatorcomment>Não entendi a diferença entre essa string e a opção &quot;Auto&quot;. Traduzir para &quot;Depende(r) do tamanho&quot; irá passar a mesma impressão do modo automático; afinal, ele também depende do tamanho disponível.</translatorcomment>
```
zaps166, line 276: Why removing comments?

zaps166 (commented):
`ru` and `uk` shouldn't be changed, too (according to other apps):
Screenshot_20220505_135404

zo-shin:
As you see in the picture, there are no ФПД on the keyboard. If such letters are used as shortcut, they definitely don’t work.
As for the comment, those languages have been ignored for a long time. If there is someone who wants to revise them, those comments won’t do much work. If you resist the clean-up, I will get them back.

zaps166:
> As you see in the picture, there are no ФПД on the keyboard. If such letters are used as shortcut, they definitely don’t work. As for the comment, those languages have been ignored for a long time. If there is someone who wants to revise them, those comments won’t do much work. If you resist the clean-up, I will get them back.

It's not true, if you change keyboard layout to e.g. Ukrainian, you can use these letters and key shortcuts are working.

lang/pt_BR.ts
zaps166: @zo-shin ping here

zaps166:
Merged from command line

Issue #610, opened 2023-05-26
VTB failure on OS X 10.9


RJVB:
As mentioned in my previous ticket, I have a long-standing fails-to-work issue with VTB in QMPlay2 on my Mac (2011 MPB 8.1 running OS X 10.9.5). This hardware acceleration does work in VLC as far as I can tell, so a priori it's not that my OS is too old (VTB was introduced in 10.8 btw).

I've been trying to trace the origin of the problem but get stuck somewhere on what appears to be a C++ problem...

In `FFDecVTP::open()`:

`av_hwdevice_ctx_create(&m_hwDeviceBufferRef, AV_HWDEVICE_TYPE_VIDEOTOOLBOX, nullptr, nullptr, 0)`

does *not* fail. The subsequent operation

`vtbOpenGL = make_shared<VTBOpenGL>(m_hwDeviceBufferRef);`

also appears to succeed (I get a `VTBOpenGL` instance with its own, non-null copy of `m_hwDeviceBufferRef`).

Things go wrong in `OpenGLWriter::setHWDecContext()`: the dynamic cast from the `HWDecContext` instance to a `OpenGLHWInterop` fails. With some trace output:
```
bool OpenGLWriter::setHWDecContext(const shared_ptr<HWDecContext> &hwDecContext)
{
    auto hwInterop = dynamic_pointer_cast<OpenGLHWInterop>(hwDecContext);
    auto hwInterop2 = dynamic_cast<OpenGLHWInterop*>(hwDecContext.get());
    qWarning() << Q_FUNC_INFO << "hwDecContext" << hwDecContext.get()
        << "hwInterop" << hwInterop.get()
        << "hwInterop2" << (void*)hwInterop2;
    if (hwDecContext && !hwInterop)
        return false;
```

I get `virtual bool OpenGLWriter::setHWDecContext(const shared_ptr<HWDecContext> &) hwDecContext 0x11e4061c8 hwInterop 0x0 hwInterop2 0x0`

As far as I can tell this makes little sense: `VTBOpenGL` inherits `OpenGLHWInterop` inherits `HWDecContext` so unless something fishy is going on with the `shared_ptr` implementation a `VTBOpenGL` instance passed as a `hwDecContext` pointer should be dynamic-castable to a `OpenGLHWInterop` pointer.

Any idea how to figure out what goes wrong and why, here?

How crucial is the use of `shared_ptr`?

RJVB:
[...]

PR #615, opened 2023-06-08
Do not attempt to use VTB for VP9 decoding when not supported

RJVB:
The VideoToolBox framework on Mac gained support for VP9 decoding somewhere in 2018, most likely with the release of Mac OS 10.15 . Only allow VTB hw-accelerated decoding of VP9 content from that OS version onwards.

QMPlay2 no longer crashes without this protection (since https://github.com/zaps166/QMPlay2/issues/610#issuecomment-1579349256?); instead there is a visible reset and most of the information in the `Information` window disappears.

NB: working VTB acceleration still requires building with default (public) C++ visibibility (or merging the FFmpeg extension into libqmplay2).

### DIFF
--- a/src/modules/FFmpeg/FFDecVTB.cpp
+++ b/src/modules/FFmpeg/FFDecVTB.cpp
@@ -75,7 +75,16 @@ bool FFDecVTB::open(StreamInfo &streamInfo)
 {
     if (streamInfo.params->codec_type != AVMEDIA_TYPE_VIDEO)
         return false;
-
+    if (streamInfo.params->codec_id == AV_CODEC_ID_VP9)
+    {
+#if __has_builtin(__builtin_available)
+        if (! __builtin_available(macOS 10.15, *))
+#endif
+        {
+            qWarning() << "VP9 not supported by VTB";
+            return false;
+        }
+    }
     const AVPixelFormat pix_fmt = streamInfo.pixelFormat();
     if (pix_fmt == AV_PIX_FMT_YUV420P10)
     {
--- a/src/modules/FFmpeg/FFDecVTB.cpp
+++ b/src/modules/FFmpeg/FFDecVTB.cpp
@@ -34,9 +34,60 @@ extern "C"
     #include <libavcodec/avcodec.h>
     #include <libavutil/hwcontext.h>
     #include <libavutil/hwcontext_videotoolbox.h>
+    #include <libavcodec/videotoolbox.h>
 }
 
+#include <dlfcn.h>
+
 using namespace std;
+
+static bool isCodecSupported(const StreamInfo &streamInfo) // Put this above "vtbGetFormat"
+{
+    CMVideoCodecType cmCType;
+    switch (streamInfo.params->codec_id)
+    {
+        case AV_CODEC_ID_H263:
+            cmCType = kCMVideoCodecType_H263;
+            break;
+        case AV_CODEC_ID_H264:
+            cmCType = kCMVideoCodecType_H264;
+            break;
+        case AV_CODEC_ID_HEVC:
+            // kCMVideoCodecType_HEVC isn't defined on all Mac OS versions
+            cmCType = 'hvc1';
+            break;
+        case AV_CODEC_ID_MPEG1VIDEO:
+            cmCType = kCMVideoCodecType_MPEG1Video;
+            break;
+        case AV_CODEC_ID_MPEG2VIDEO:
+            cmCType = kCMVideoCodecType_MPEG2Video;
+            break;
+        case AV_CODEC_ID_MPEG4:
+            cmCType = kCMVideoCodecType_MPEG4Video;
+            break;
+        case AV_CODEC_ID_VP9:
+            // kCMVideoCodecType_VP9 isn't defined on all Mac OS versions
+            cmCType = 'vp09';
+            break;
+        default:
+            cmCType = 0;
+            break;
+    }
+
+    if (!cmCType)
+        return false;
+
+    // VTIsHardwareDecodeSupported() was introduced in 10.13 only so in order to run on older OS versions
+    // without resorting to conditional code we obtain (and cache) a pointer to the function via dlsym().
+    // According to https://www.objc.io/issues/23-video/videotoolbox/ Macs (running OS X 10.10) support
+    // "usually both H.264 and MPEG-4 Part 2 in hardware". H263 and MPEG 1,2 are supported in software
+    // which is of no use for us here.
+    static auto VTIsHardwareDecodeSupported = (bool (*)(CMVideoCodecType))dlsym(RTLD_DEFAULT, "VTIsHardwareDecodeSupported");
+    return VTIsHardwareDecodeSupported 
+        ? VTIsHardwareDecodeSupported(cmCType)
+        : (cmCType == kCMVideoCodecType_H264 || cmCType == kCMVideoCodecType_MPEG4Video)
+    ;
+}
 
 static AVPixelFormat vtbGetFormat(AVCodecContext *codecCtx, const AVPixelFormat *pixFmt)
 {
@@ -73,18 +124,9 @@ QString FFDecVTB::name() const
 
 bool FFDecVTB::open(StreamInfo &streamInfo)
 {
-    if (streamInfo.params->codec_type != AVMEDIA_TYPE_VIDEO)
+    if (streamInfo.params->codec_type != AVMEDIA_TYPE_VIDEO || !hasHWAccel("videotoolbox"))
         return false;
-    if (streamInfo.params->codec_id == AV_CODEC_ID_VP9)
-    {
-#if __has_builtin(__builtin_available)
-        if (! __builtin_available(macOS 10.15, *))
-#endif
-        {
-            qWarning() << "VP9 not supported by VTB";
-            return false;
-        }
-    }
+
     const AVPixelFormat pix_fmt = streamInfo.pixelFormat();
     if (pix_fmt == AV_PIX_FMT_YUV420P10)
     {
@@ -95,10 +137,19 @@ bool FFDecVTB::open(StreamInfo &streamInfo)
     {
         return false;
     }
+    
+    if (!isCodecSupported(streamInfo))
+    {
+        qWarning() << streamInfo.codec_name << "is not supported by VTB";
+        return false;
+    }
 
     AVCodec *codec = init(streamInfo);
-    if (!codec || !hasHWAccel("videotoolbox"))
+    if (!codec)
+    {
+        qWarning() << "VTB: no or unsupported codec";
         return false;
+    }
 
 #ifdef USE_OPENGL
     shared_ptr<VTBOpenGL> vtbOpenGL;
@@ -112,14 +163,20 @@ bool FFDecVTB::open(StreamInfo &streamInfo)
 #endif
 
     if (!m_hwDeviceBufferRef && av_hwdevice_ctx_create(&m_hwDeviceBufferRef, AV_HWDEVICE_TYPE_VIDEOTOOLBOX, nullptr, nullptr, 0) != 0)
+    {
+        qWarning() << "VTB: failed to create hwdevice_ctx";
         return false;
+    }
 
 #ifdef USE_OPENGL
     if (QMPlay2Core.renderer() == QMPlay2CoreClass::Renderer::OpenGL && !vtbOpenGL)
     {
         vtbOpenGL = make_shared<VTBOpenGL>(m_hwDeviceBufferRef);
         if (!QMPlay2Core.gpuInstance()->setHWDecContextForVideoOutput(vtbOpenGL))
+        {
+            qWarning() << "VTB: failed to set VTB GPU context";
             return false;
+        }
     }
 
     if (vtbOpenGL)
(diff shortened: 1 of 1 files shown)


zaps166:
Thanks for the PR, but could you check [this](https://developer.apple.com/documentation/videotoolbox/2887343-vtishardwaredecodesupported?language=objc)? It's available since macOS 10.13, it can check if codec is supported. Looks like it's FOURCC `'vp09'` string to check it. Here's Chromium [example](https://chromium.googlesource.com/chromium/src/+/master/media/gpu/mac/vt_video_decode_accelerator_mac.cc#2451). And [here](https://github.com/FFmpeg/FFmpeg/blob/024c30aa3b3cdb7d32fa5d79b0973ab8f3065ead/libavcodec/videotoolbox.c#L54) is definition.

Btw. is HEVC working correctly?

zaps166:
No built-in available - don't use VP9 (use your `ifdef` maybe). If Has built-in - check for macOS version: >= 10.13 check for codecs, else don't allow VP9.

> Any tests of this on YT?

YT doesn't use HEVC. Try [this](https://filesamples.com/formats/hevc).

RJVB:
Or I could write an overload for `VTIsHardwareDecodeSupported` (using `dlsym(RTLD_DEFAULT,VTIsHardwareDecodeSupported)` or probably even `&VTIsHardwareDecodeSupported`) which will just return an appropriate hardwired result if `VTIsHardwareDecodeSupported` isn't available.

BTW, does https://filesamples.com/samples/video/hevc/sample_1280x720_surfing_with_audio.hevc really contain audio as the name suggests?

zaps166:
In VAAPI I'm checking for supported codecs by using VAAPI functions. It'll be good to do it for VTB, too (within an `ifdef`).

RJVB:
>It's possible to resolve symbol, but it might be too complicated. It's backwards compat, let it just "work".

Not sure what you mean with that, but this wouldn't be my first rodeo to do this kind of providing a wrapper for a possibly missing "system" function. I could even make it really fancy by fetching the actual VTB function in a shared library initialisation function but there would not be a real performance benefit to that.

>Btw what happens now with current master without this patch? Does it still crashes?

That's described in the "OP" ;) You get a bit of an ugly-looking reset/restart, and most of the information in the `Information` panel gets lost. Interestingly neither of those occur when the hwaccel context is reset after playing a supported video with hw acceleration.

zaps166:
> That's described in the "OP" ;)

xD

> and most of the information in the `Information` panel gets lost

That's interesting. Screenshot?

> You get a bit of an ugly-looking reset/restart

I'm aware it'll switch to software decoder, but it should be quick. That's why better to check it using API :slightly_smiling_face:

RJVB:
> I'm aware it'll switch to software decoder, but it should be quick. That's why better to check it using API 🙂

Agreed. Now the question is, how are we going to map the AVCodecID to a CMVideoCodecType? FFmpeg 4.4.4 doesn't support decoding content with VTB for all codecs that VTB supports, fortunately, because it doesn't export a dedicated mapping function that you could call without going through an entire context initialisation. 
In `libavcodec/videotoolbox.c` I'm seeing evidence of
kCMVideoCodecType_HEVC
kCMVideoCodecType_MPEG4Video
kCMVideoCodecType_H264
kCMVideoCodecType_H263
kCMVideoCodecType_MPEG1Video
kCMVideoCodecType_MPEG2Video

`libavcodec/videotoolboxenc.c` also has kCMVideoCodecType_HEVCWithAlpha but I guess that's for encoding only(?).

RJVB:
BTW, I appear not to have HEVC hw accelerated decoding either. Interestingly, when I open the above video via its link in QMPlay2 it plays, but apparently fast-as-you-can. After downloading it QMPlay2 opens it and stops immediately. `ffplay` (from the same FFmpeg) and VLC (official build) both play the video at the correct rate.

zaps166:
> Now the question is, how are we going to map the AVCodecID to a CMVideoCodecType?

[By name](https://github.com/zaps166/QMPlay2/blob/e1dfb228d10a7d79072e4d2f8c18eba92168abeb/src/modules/FFmpeg/VAAPI.cpp#L415) :smile: Or by `AVCodecID` in `switch`.

> QMPlay2 it plays, but apparently fast-as-you-can.

Using VTB or software decoder?

RJVB:
> Using VTB or software decoder?

It looks like FFmpeg assumes that VTB support is available, and maybe it is to some extent. I dunno how to force ffplay to use hwaccel so I can't check that but apparently it's FFmpeg that goes wrong here somewhere. Curiously I didn't see the VTB tag in the info window.

I now have a working `isSupported()` function, returning false for HEVC, and with that playback is at the appropriate speed.

```
static bool isSupported(StreamInfo &streamInfo)
{
    static bool (*VTIsHardwareDecodeSupported)(CMVideoCodecType) = nullptr;
    if (!VTIsHardwareDecodeSupported)
    {
        VTIsHardwareDecodeSupported = (bool (*)(CMVideoCodecType)) dlsym(RTLD_DEFAULT, "VTIsHardwareDecodeSupported");
    }
    if (streamInfo.params->codec_type != AVMEDIA_TYPE_VIDEO)
        return false;
    CMVideoCodecType cmCType;
    switch (streamInfo.params->codec_id)
    {
        case AV_CODEC_ID_H263:
            cmCType = kCMVideoCodecType_H263;
            break;
        case AV_CODEC_ID_H264:
            cmCType = kCMVideoCodecType_H264;
            break;
        case AV_CODEC_ID_HEVC:
            // kCMVideoCodecType_HEVC isn't defined on all Mac OS versions
            cmCType = 'hvc1';
            break;
        case AV_CODEC_ID_MPEG1VIDEO:
            cmCType = kCMVideoCodecType_MPEG1Video;
            break;
        case AV_CODEC_ID_MPEG2VIDEO:
            cmCType = kCMVideoCodecType_MPEG2Video;
            break;
        case AV_CODEC_ID_MPEG4:
            cmCType = kCMVideoCodecType_MPEG4Video;
            break;
        case AV_CODEC_ID_VP9:
            // kCMVideoCodecType_VP9 isn't defined on all Mac OS versions
            cmCType = 'vp09';
        default :
            cmCType = 0;
    }
    bool supported = false;
    if (cmCType)
    {
        if (VTIsHardwareDecodeSupported)
        {
            supported = VTIsHardwareDecodeSupported(cmCType);
        }
        else
        {
            // we'll need to do some guessing here, so let's assume that all of
            // the above will work except for HEVC and VP9
            supported = (cmCType != 'vp09' && cmCType != 'hvc1');
        }
    }
        
    if (!supported)
    {
        qWarning() << "codec not supported by VTB";
        return false;
    }
    const AVPixelFormat pix_fmt = streamInfo.pixelFormat();
    if (pix_fmt == AV_PIX_FMT_YUV420P10)
    {
        if (streamInfo.params->codec_id == AV_CODEC_ID_H264)
        {
            qWarning() << "H264/yuv420p10 not supported by VTB";
            return false;
        }
    }
    else if (pix_fmt != AV_PIX_FMT_YUV420P && pix_fmt != AV_PIX_FMT_YUVJ420P)
    {
        return false;
    }
    return FFDecHWAccel::hasHWAccel("videotoolbox");
}

bool FFDecVTB::open(StreamInfo &streamInfo)
{
    if (!isSupported(streamInfo))
    {
        return false;
    }

    AVCodec *codec = init(streamInfo);
    if (!codec)
    {
        qWarning() << "VTB: no or unsupported codec";
        return false;
    }

```

zaps166:
After quick refactor (not checked at all):
```c++
static bool isCodecSupported(const StreamInfo &streamInfo) // Put this above "vtbGetFormat"
{
    CMVideoCodecType cmCType;
    switch (streamInfo.params->codec_id)
    {
        case AV_CODEC_ID_H263:
            cmCType = kCMVideoCodecType_H263;
            break;
        case AV_CODEC_ID_H264:
            cmCType = kCMVideoCodecType_H264;
            break;
        case AV_CODEC_ID_HEVC:
            // kCMVideoCodecType_HEVC isn't defined on all Mac OS versions
            cmCType = 'hvc1';
            break;
        case AV_CODEC_ID_MPEG1VIDEO:
            cmCType = kCMVideoCodecType_MPEG1Video;
            break;
        case AV_CODEC_ID_MPEG2VIDEO:
            cmCType = kCMVideoCodecType_MPEG2Video;
            break;
        case AV_CODEC_ID_MPEG4:
            cmCType = kCMVideoCodecType_MPEG4Video;
            break;
        case AV_CODEC_ID_VP9:
            // kCMVideoCodecType_VP9 isn't defined on all Mac OS versions
            cmCType = 'vp09';
            break;
        default:
            cmCType = 0;
            break;
    }

    if (!cmCType)
        return false;

    static auto VTIsHardwareDecodeSupported = (bool (*)(CMVideoCodecType))dlsym(RTLD_DEFAULT, "VTIsHardwareDecodeSupported");
    return VTIsHardwareDecodeSupported 
        ? VTIsHardwareDecodeSupported(cmCType)
        : (cmCType != 'vp09' && cmCType != 'hvc1') // let's assume that all of the above will work except for HEVC and VP9
    ;
}

...

bool FFDecVTB::open(StreamInfo &streamInfo)
{
    if (streamInfo.params->codec_type != AVMEDIA_TYPE_VIDEO)
        return false;

    const AVPixelFormat pix_fmt = streamInfo.pixelFormat();
    if (pix_fmt == AV_PIX_FMT_YUV420P10)
    {
        if (streamInfo.params->codec_id == AV_CODEC_ID_H264)
            return false;
    }
    else if (pix_fmt != AV_PIX_FMT_YUV420P && pix_fmt != AV_PIX_FMT_YUVJ420P)
    {
        return false;
    }
    
    if (!isCodecSupported(streamInfo))
    {
        qWarning() << streamInfo.codec_name << "is not supported by VTB";
        return false;
    }
    
    ...
```

`FFDecHWAccel::hasHWAccel()` good, but in this case it's good to change in all files to call `FFDecHWAccel::hasHWAccel` and it can be called earlier. Just different task :smile:

RJVB:
>static bool isCodecSupported(const StreamInfo &streamInfo) // Put this above "vtbGetFormat"
[...]
>bool FFDecVTB::open(StreamInfo &streamInfo)
>{
>    if (streamInfo.params->codec_type != AVMEDIA_TYPE_VIDEO)
>        return false;
>
>    const AVPixelFormat pix_fmt = streamInfo.pixelFormat();

I thought it clearer to put all the "is this one supported" logic in a single function. Why do you prefer this?

`hasHWAccel()` doesn't have to be called with the class prefixed but isn't that sort of cosmetic change something for you to do anyway? ;)

zaps166:
> `hasHWAccel()` doesn't have to be called with the class prefixed but isn't that sort of cosmetic change something for you to do anyway? ;)

Ah yes, it's in the same class :sweat_smile: It can be `static`, also it can be called before `init()` everywhere (and before `isCodecSupported()` here).

> I thought it clearer to put all the "is this one supported" logic in a single function. Why do you prefer this?

I think it'll be consistent with all other HW decoders + less changes. There will be just a separate function which calls VTB outside FFmpeg to check for supported codec.

RJVB:
>Ah yes, it's in the same class :sweat_smile: It can be `static`, also it can be called before `init()` everywhere (and before `isCodecSupported()` here).

It would probably make even more sense if this cached the result of the iteration. I doubt support for hw acceleration in general (or VTB specifically) will vary between calls ;)

zaps166:
> I doubt support for hw acceleration in general (or VTB specifically) will vary between calls ;)

It will not, but I guess checking is about few microseconds and it's done only once when opening video :sweat_smile:

zaps166:
Check this a few times:
```c++
QElapsedTimer et;
et.start();
VTIsHardwareDecodeSupported(...);
auto t = et.nsecsElapsed();
qDebug() << t / 1e3 << "us";
```

RJVB:
Oh, I don't doubt there is little if any performance gain to be had here ;)

src/modules/FFmpeg/FFDecVTB.cpp
```diff
@@ -34,10 +34,61 @@ extern "C"
...
     #include <libavutil/hwcontext.h>
     #include <libavutil/hwcontext_videotoolbox.h>
+    #include <libavcodec/videotoolbox.h>
 }
 
+#include <dlfcn.h>
+
 using namespace std;
 
+static bool isCodecSupported(const StreamInfo &streamInfo) // Put this above "vtbGetFormat"
```
zaps166, line 44: It was comment to you where to put XD

RJVB:
>It was comment to you where to put XD

Oops, did I leave it in? I guess I wanted to see if you'd notice O:^)

BTW, I was a bit surprised the static variable initialisation with a call to dlsym() worked, as the VTB framework isn't linked in directly. Something to keep in the back of mind.

zaps166:
> Oops, did I leave it in? I guess I wanted to see if you'd notice O:^)

:sweat_smile: 

> BTW, I was a bit surprised the static variable initialisation with a call to dlsym() worked

It's initialized on first call, next time it's already initialized, so it will not call `dlsym` anymore. It should be thread-safe in this case in modern C++, but your previous code was not thread-safe.
`dlsym` with `RTLD_DEFAULT` looks also for symbols in dependencies, I guess FFmpeg links to it or loads it.

RJVB:
I was more concerned that the library might not yet have been loaded, as it's not linked directly. It should evidently be available when the function is called.

Is being thread-safe a consideration here? The function cannot be called from multiple threads at the same time, right?

PR #618, opened 2023-06-18
Make classes required for VideoToolBox support visible across modules

### LINKED ISSUE #610: VTB failure on OS X 10.9

RJVB:
VideoToolBox support failed on certain older Mac OS X versions because hidden visibility caused a dynamic cast from a FFDecVTB instance to fail. This issue is fixed by exporting the classes involved.

Fixes #610

### DIFF
--- a/src/modules/FFmpeg/FFDecVTB.hpp
+++ b/src/modules/FFmpeg/FFDecVTB.hpp
@@ -20,7 +20,7 @@
 
 #include <FFDecHWAccel.hpp>
 
-class FFDecVTB final : public FFDecHWAccel
+class FFMPEGVTB_EXPORT FFDecVTB final : public FFDecHWAccel
 {
 public:
     FFDecVTB(Module &module);
--- a/src/modules/FFmpeg/CMakeLists.txt
+++ b/src/modules/FFmpeg/CMakeLists.txt
@@ -127,6 +127,8 @@ if(USE_FFMPEG_VAAPI OR USE_FFMPEG_VDPAU OR USE_FFMPEG_DXVA2 OR USE_FFMPEG_D3D11V
     endif()
 endif()
 
+add_definitions(-DFFMPEG_MODULE)
+
 add_library(${PROJECT_NAME} ${QMPLAY2_MODULE}
     ${FFmpeg_HDR}
     ${FFmpeg_SRC}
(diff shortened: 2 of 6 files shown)


zaps166:
Thanks!

src/modules/FFmpeg/FFDecVTB.hpp
RJVB, line 1: You forgot to change the macro here, in #50d8960a70ef8e52b46e91a9038b4ef369575976 ! ;)

PR #666, opened 2023-10-29
Support reading lyrics meta-data from ID3v2 tags

Martchus:
ID3v2 tags support specifying a description and the language of lyrics. There can be one lyric per description/language combination. Hence libavformat makes up keys like `lyrics-foo-eng` in the metadata dictionary for those fields (see code in `libavformat/id3v2.c` in the FFmpeg repository or call `ffprobe` on such a file). Even if the description and language are empty the field name for lyrics becomes `lyrics-`. This change allows QMPlay2 to display the first of these fields.

### DIFF
--- a/src/modules/FFmpeg/FormatContext.cpp
+++ b/src/modules/FFmpeg/FormatContext.cpp
@@ -25,6 +25,7 @@
 #include <Settings.hpp>
 #include <Packet.hpp>
 
+#include <cstring>
 #include <limits>
 
 extern "C"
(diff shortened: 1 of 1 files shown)


Martchus:
And yes, this introduces a slow linear lookup. I think this isn't a big deal for just one field considering the usually small size of the dict. If you think it is too much we could also just introduce a lookup for `lyrics-` which should cover most cases (who tags the language/description anyways).

---

One could also display all the lyrics like this:

```
description 1 (language 1):
the lyrics 1

description 2 (language 2):
the lyrics 2
```

Of course it would be a little bit more code considering one had to handle the case when either description or language are empty and turn the abbreviated language into a displayable string if possible (e.g. via `QLocale`).

zaps166:
Thanks!

Martchus:
Also thanks for merging as-is :-)

zaps166:
I didn't test it 😅

PR #693, opened 2024-04-06
Mark error logs untranslatable

yakushabb:
We don't need to mark error logs as translatable.


zaps166:
What happened with https://github.com/zaps166/QMPlay2/pull/660 ?

yakushabb:
It was closed automatically when I deleted my old github account.

zaps166:
Errors displayed in status bar should be translated.

yakushabb:
Which ones displayed in the status bar? I assumed they are just log entires.

zaps166:
`logInfo` and `logError` has `showInGUI` argument (which is `true` by default), so most (all?) of them.

yakushabb:
In that case, may I merge splitted sentences?

https://wiki.gnome.org/TranslationProject/DevGuidelines/Never%20split%20sentences

zaps166:
> In that case, may I merge splitted sentences?

It requires to retranslate many sentences for all languages. But yes, try it, we'll see!

yakushabb:
I will open another PR soon. Thanks.

PR #736, opened 2024-06-09
Update RadioBrowser API URL

zo-shin:
Update RadioBrowser API URL


zaps166:
Is `de1` server always active?

zo-shin:
Yes, it’s one of the available servers.

zaps166:
I remember the recommended solution is different (DNS lookup). I'll keep it as-is for now as long as it works. Thanks anyway!

PR #747, opened 2024-07-06
Fix Strings Disorder

zo-shin:
Fix tr strings disorder


zaps166:
Thanks, haven't checked it yet!

PR #834, opened 2024-12-27
Simplify Strings

zo-shin:
simplify strings

### DIFF
--- a/lang/ar.ts
+++ b/lang/ar.ts
@@ -1926,8 +1926,8 @@
         <translation>مرشحات صوت FFmpeg، مثال: %1</translation>
     </message>
     <message>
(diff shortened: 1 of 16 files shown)


zaps166:
But there are two options which have to be used :thinking: :
Zrzut ekranu_20241227_172354

zo-shin:
How about the new one? Or do you have any better idea?

zaps166:
Something like this: ", you need to enable channel conversion and use it before filters option in playback settings.", but it's so hard to explain :smile:

zaps166:
Or maybe better: ", you need to enable the channel conversion and enable option to use channel conversion before filters in playback settings." ?

zo-shin:
quotes add

zaps166:
Thanks!

PR #859, opened 2025-01-17
YouTubeDL: Allow passing HTTP referrer in additional module parameters

Realex-fire:
Hello @zaps166,
You have already allowed cookies from browsers (2c9d7c2), but sometimes that is not enough and you can still get a 403 error.
Some network streams require custom HTTP headers and this looks redundant for the media player. But the HTTP referrer can be easily extracted and can be used in additional module parameters.

This PR allows it and neatly extends the current QMPlay2 behavior.
I hope this will be useful. What do you think?

### DIFF
--- a/src/qmplay2/YouTubeDL.cpp
+++ b/src/qmplay2/YouTubeDL.cpp
@@ -124,7 +124,7 @@ void YouTubeDL::addr(const QString &url, const QString &param, QString *streamUr
     QStringList paramList {"-e"};
     if (!param.isEmpty())
     {
-        paramList << "-f" << param;
+        paramList << (QUrl(param).host().isEmpty() ? "-f" : "--referer") << param;
     }
     else
     {


zaps166:
- it's only for yt-dlp
- it replaces the `-f` - is it good to use it instead?

zaps166:
> Yes, for ffmpeg you have [different](https://github.com/zaps166/QMPlay2/blob/2a36237428f0a9180032345bf6fecfc3c77f4b75/src/modules/FFmpeg/FFDemux.cpp#L258) and more complex [logic](https://github.com/zaps166/QMPlay2/blob/2a36237428f0a9180032345bf6fecfc3c77f4b75/src/modules/FFmpeg/FormatContext.cpp#L705).

`referer` to FFmpeg can be added by `QMPlay2CoreClass::addRawHeaders()` per URL. But I don't have any URL to test, I never needed it yet.

> yt-dlp should be enough.

As long as it works, I'll merge it, thanks! But one question - what if we always add `referer` from `url` argument there instead?

zaps166:
> You can try to extract any stream from [this](https://www.earthcam.com/) website, you need to use this domain as referer.

`yt-dlp` extracted video, but it doesn't play. Also I can't play it when I set same `referer` to FFmpeg. Any ideas?

zaps166:
Currently I have this:
```diff
diff --git a/src/qmplay2/YouTubeDL.cpp b/src/qmplay2/YouTubeDL.cpp
index b19c8990..5bf1d443 100644
--- a/src/qmplay2/YouTubeDL.cpp
+++ b/src/qmplay2/YouTubeDL.cpp
@@ -121,10 +121,16 @@ void YouTubeDL::addr(const QString &url, const QString &param, QString *streamUr
     if (!streamUrl && !name)
         return;
 
+    QString referer;
+
     QStringList paramList {"-e"};
     if (!param.isEmpty())
     {
-        paramList << "-f" << param;
+        if (!QUrl(param).host().isEmpty())
+        {
+            referer = param;
+        }
+        paramList << (referer.isEmpty() ? "-f" : "--referer") << param;
     }
     else
     {
@@ -154,6 +160,11 @@ void YouTubeDL::addr(const QString &url, const QString &param, QString *streamUr
                 *streamUrl += "[" + tmpUrl + "]";
             *streamUrl += "}";
         }
+        if (!referer.isEmpty())
+        {
+            for (const QString &url : std::as_const(ytdlStdout))
+                QMPlay2Core.addRawHeaders(url, "Referer: " + referer.toUtf8() + "\r\n");
+        }
     }
     if (name && !title.isEmpty())
         *name = title;
```

zaps166:
> I checked these changes, everything works for me. Maybe you have a problem with the decoder or the stream is broken?

Extracted URL when trying to play it: `HTTP error 404 Not Found`

zaps166:
Looks like yt-dlp extracts wrong URL

zaps166:
> Yes, you need a URL to the stream, not a player page. I extracted one [URL](https://videos-3.earthcam.com/fecnetwork/14013.flv/playlist.m3u8) for you.

Yes, it works, but `yt-dlp` doesn't make any sense - it's just a workaround here :sweat_smile:  Better to add `referer` to FFmpeg, too...

zaps166:
This can go, in your case, it's just a workaround for missing parameter for direct URL.

zaps166:
Are there any cases that server needs `referer` to stream (not to playlist)? In this case this will not work.

zaps166:
> It works fine with my private stream where `referer` is needed to download `.ts` fragments.

If it works to you, I'll to it :smile: 

---

Next step should be add referer to direct address to skip yt-dlp, maybe some day (needs UI changes, etc.)...

PR #968, opened 2025-09-16
Filling in gaps in Hungarian language files

Impostor0729:
Filling in gaps in Hungarian language file. 

I did not modify the existing lines, I only translated the missing lines.


zaps166:
Thanks!

PR #978, opened 2025-10-26
fix The problem with the file path led to the compilation failure

tingzhong666:
fix The problem with the file path led to the compilation failure
修复 头文件路径的问题，导致编译失败


zaps166:
This will break compilation pf `.tar.xz` package. To use git on Windows, enable symlinks (GIT on Windows should give option to enable it when installing and give instruction how to give symlink permissions on Windows, otherwise read https://stackoverflow.com/a/59761201).

PR #1004, opened 2026-04-18
QMPlay2 Update ru.ts

Vanadium23-tor:
I installed QMPlay2, but some lines weren't translated into Russian. I fixed it and translated them.


zaps166:
Thanks!

PR #1028, opened 2026-07-18
Try not to access YT streams before they're available

### LINKED ISSUE #1027: Why does this YT video refuse to play?
https://www.youtube.com/watch?v=PqML4JYv_24

When I configure QMPlay2 to invoke yt-dlp with cookies from Firefox, I get two 403 "Forbidden" errors, but when I don't use cookies the video plays back just fine. (That probably explains why `mpv` also plays it OK.)

Using the yt-dlp from the QMPlay2 folder, I get this table without cookies:
```
[youtube] Extracting URL: https://www.youtube.com/watch?v=PqML4JYv_24
[youtube] PqML4JYv_24: Downloading webpage
[youtube] PqML4JYv_24: Downloading android vr player API JSON
[info] Available formats for PqML4JYv_24:
ID  EXT   RESOLUTION FPS CH │   FILESIZE   TBR PROTO │ VCODEC        VBR ACODEC      ABR ASR MORE INFO
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────                                                                                                       
sb3 mhtml 48x27        1    │                  mhtml │ images                                storyboard
sb2 mhtml 80x45        1    │                  mhtml │ images                                storyboard
sb1 mhtml 160x90       1    │                  mhtml │ images                                storyboard
sb0 mhtml 320x180      1    │                  mhtml │ images                                storyboard
139 m4a   audio only      2 │  732.21KiB   49k https │ audio only        mp4a.40.5   49k 22k [en-US] English (US) original (default), low, m4a_dash
140 m4a   audio only      2 │    1.89MiB  130k https │ audio only        mp4a.40.2  130k 44k [en-US] English (US) original (default), medium, m4a_dash
251 webm  audio only      2 │    1.94MiB  133k https │ audio only        opus       133k 48k [en-US] English (US) original (default), medium, webm_dash
160 mp4   256x144     30    │    1.42MiB   97k https │ avc1.4d400c   97k video only          144p, mp4_dash
134 mp4   640x360     30    │    7.49MiB  513k https │ avc1.4d401e  513k video only          360p, mp4_dash
18  mp4   640x360     30  2 │    8.19MiB  559k https │ avc1.42001E       mp4a.40.2       22k [en] 360p
298 mp4   1280x720    60    │   49.46MiB 3383k https │ avc1.640020 3383k video only          720p60, mp4_dash
299 mp4   1920x1080   60    │   87.82MiB 6007k https │ avc1.64002a 6007k video only          1080p60, mp4_dash
```

and this *with* cookies:

```
Extracting cookies from firefox
Extracted 911 cookies from firefox
[youtube] Extracting URL: https://www.youtube.com/watch?v=PqML4JYv_24
[youtube] PqML4JYv_24: Downloading webpage
[youtube] PqML4JYv_24: Downloading tv downgraded player API JSON
[youtube] PqML4JYv_24: Downloading player ecd4b80a-main
[youtube] [jsc:node] Solving JS challenges using node
[youtube] PqML4JYv_24: Downloading m3u8 information
[info] Available formats for PqML4JYv_24:
ID  EXT   RESOLUTION FPS CH │   FILESIZE   TBR PROTO │ VCODEC        VBR ACODEC      ABR ASR MORE INFO
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
sb3 mhtml 48x27        1    │                  mhtml │ images                                storyboard
sb2 mhtml 80x45        1    │                  mhtml │ images                                storyboard
sb1 mhtml 160x90       1    │                  mhtml │ images                                storyboard
sb0 mhtml 320x180      1    │                  mhtml │ images                                storyboard
140 m4a   audio only      2 │    1.89MiB  130k https │ audio only        mp4a.40.2  130k 44k [en-US] English (US) original (default), medium, m4a_dash
251 webm  audio only      2 │    1.94MiB  133k https │ audio only        opus       133k 48k [en-US] English (US) original (default), medium, webm_dash
91  mp4   256x144     30    │ ~  2.72MiB  186k m3u8  │ avc1.4D400C       mp4a.40.5           [en-US] (original)
160 mp4   256x144     30    │    1.42MiB   97k https │ avc1.4d400c   97k video only          144p, mp4_dash
93  mp4   640x360     30    │ ~ 12.03MiB  820k m3u8  │ avc1.4D401E       mp4a.40.2           [en-US] (original)
134 mp4   640x360     30    │    7.49MiB  513k https │ avc1.4d401e  513k video only          360p, mp4_dash
18  mp4   640x360     30  2 │    8.19MiB  559k https │ avc1.42001E       mp4a.40.2       22k [en] 360p
243 webm  640x360     30    │    6.34MiB  434k https │ vp9          434k video only          360p, webm_dash
136 mp4   1280x720    30    │   31.36MiB 2145k https │ avc1.64001f 2145k video only          720p, mp4_dash
300 mp4   1280x720    60    │ ~ 65.88MiB 4493k m3u8  │ avc1.640020       mp4a.40.2           [en-US] (original)
298 mp4   1280x720    60    │   49.46MiB 3383k https │ avc1.640020 3383k video only          720p60, mp4_dash
301 mp4   1920x1080   60    │ ~104.47MiB 7125k m3u8  │ avc1.64002A       mp4a.40.2           [en-US] (original)
299 mp4   1920x1080   60    │   87.82MiB 6007k https │ avc1.64002a 6007k video only          1080p60, mp4_dash
```

I usually contend myself with 720p video, and manual download works as it should:

```
> ~/.qmplay2/yt-dlp  --cookies-from-browser firefox -f 298,251 "https://www.youtube.com/watch?v=PqML4JYv_24"
Extracting cookies from firefox
Extracted 911 cookies from firefox
[youtube] Extracting URL: https://www.youtube.com/watch?v=PqML4JYv_24
[youtube] PqML4JYv_24: Downloading webpage
[youtube] PqML4JYv_24: Downloading tv downgraded player API JSON
[youtube] PqML4JYv_24: Downloading player ecd4b80a-main
[youtube] [jsc:node] Solving JS challenges using node
[youtube] PqML4JYv_24: Downloading m3u8 information
[info] PqML4JYv_24: Downloading 2 format(s): 298, 251
[download] Sleeping 4.00 seconds as required by the site...
[download] Destination: Flexi-Hose Expanding Watering Hose Review After 6 Months Use [PqML4JYv_24].mp4
[download] 100% of   49.46MiB in 00:00:01 at 26.17MiB/s
[download] Destination: Flexi-Hose Expanding Watering Hose Review After 6 Months Use [PqML4JYv_24].webm
[download] 100% of    1.94MiB in 00:00:00 at 12.16MiB/s
```

(I'll check if my recent hack to run 2 yt_dlp processes in parallel helps here, one with and one without cookies, and then use the latter as a fallback if the first fails.)

RJVB:
YouTube (now?) provides an `available_at` timestamp for streams, and it seems that accessing them before that time can result in a 403 Forbidden error.

This commit introduces a simple fix that determines which of the streams of a given video is available last, and then does a non- blocking wait before attempting to open the selected streams. The wait is signalled via a statusbar message.

There are no side-effects other than that the wait is imposed also when pasting a YT video link directly into the playlist, apparently because this operation involves starting playback. This should be avoidable but will require more work.

#Closes: #1027

### DIFF
--- a/src/modules/Extensions/YouTube.cpp
+++ b/src/modules/Extensions/YouTube.cpp
@@ -20,6 +20,7 @@
 
 #include <YouTubeDL.hpp>
 #include <LineEdit.hpp>
+#include <Functions.hpp>
 
 #include <QLoggingCategory>
 #include <QStringListModel>
@@ -46,7 +47,6 @@
 
 #include <QDateTime>
 #include <QDeadlineTimer>
-#include <unistd.h>
 
 Q_LOGGING_CATEGORY(youtube, "Extensions/YouTube")
 
@@ -1304,9 +1304,9 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
                 urlLanguages[url] = format[QStringLiteral("language")].toString();
                 urlNotes[url] = note;
             }
-            if (format.contains("available_at"))
+            if (format.contains(QStringLiteral("available_at")))
             {
-                auto available_at = format["available_at"].toDouble() * 1000.0;
+                auto available_at = format[QStringLiteral("available_at")].toDouble() * 1000.0;
                 qCDebug(youtube) << "url for format" << format["format_id"].toString() << "will be available at"
                     << QDateTime::fromMSecsSinceEpoch(available_at) << "so in" << (available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0 << "s";
                 if (available_at > latest_available_at)
@@ -1454,22 +1454,24 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
 
     if (latest_available_at > 0)
     {
-        const auto dt = int( (latest_available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0 + 0.5);
+        // The number of whole seconds until the "slowest" stream is promised to be available, plus 1
+        // This interval can be <=0 if the stream is already available!
+        const auto dt = int( (latest_available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0) + 1;
         if (dt > 0) {
             QMPlay2Core.logInfo(QString::asprintf("Waiting for %ds as required by YouTube", dt));
             // do an active wait; a less active clone of QTest::qWait():
-            auto remaining = dt * 1000;
+            auto remaining = dt * 1000.0;
             QDeadlineTimer timer(remaining, Qt::PreciseTimer);
             do
             {
                 QCoreApplication::processEvents(QEventLoop::AllEvents, remaining);
-                QCoreApplication::sendPostedEvents(Q_NULLPTR, QEvent::DeferredDelete);
+                QCoreApplication::sendPostedEvents(nullptr, QEvent::DeferredDelete);
                 remaining = timer.remainingTime();
                 if (remaining <= 0)
                 {
                     break;
                 }
-                usleep(qMin(100, remaining) * 1000);
+                Functions::s_wait(qMin(100.0, remaining) / 1000.0);
                 remaining = timer.remainingTime();
             } while (remaining > 0);
         }
--- a/src/modules/Extensions/YouTube.cpp
+++ b/src/modules/Extensions/YouTube.cpp
@@ -44,6 +44,10 @@
 #include <QMenu>
 #include <QUrl>
 
+#include <QDateTime>
+#include <QDeadlineTimer>
+#include <unistd.h>
+
 Q_LOGGING_CATEGORY(youtube, "Extensions/YouTube")
 
 #define YOUTUBE_URL "https://www.youtube.com"
@@ -1271,6 +1275,8 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
 
     QHash<int, QPair<QStringList, QStringList>> itagsData;
 
+    double latest_available_at = 0.0;
+
     for (auto &&formatVal : formats)
     {
         const auto format = formatVal.toObject();
@@ -1297,6 +1303,16 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
             {
                 urlLanguages[url] = format[QStringLiteral("language")].toString();
                 urlNotes[url] = note;
+            }
+            if (format.contains("available_at"))
+            {
+                auto available_at = format["available_at"].toDouble() * 1000.0;
+                qCDebug(youtube) << "url for format" << format["format_id"].toString() << "will be available at"
+                    << QDateTime::fromMSecsSinceEpoch(available_at) << "so in" << (available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0 << "s";
+                if (available_at > latest_available_at)
+                {
+                    latest_available_at = available_at;
+                }
             }
         }
     }
@@ -1436,6 +1452,28 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
 
     result += o["description"].toString();
 
+    if (latest_available_at > 0)
+    {
+        const auto dt = int( (latest_available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0 + 0.5);
+        if (dt > 0) {
+            QMPlay2Core.logInfo(QString::asprintf("Waiting for %ds as required by YouTube", dt));
+            // do an active wait; a less active clone of QTest::qWait():
+            auto remaining = dt * 1000;
+            QDeadlineTimer timer(remaining, Qt::PreciseTimer);
+            do
+            {
+                QCoreApplication::processEvents(QEventLoop::AllEvents, remaining);
+                QCoreApplication::sendPostedEvents(Q_NULLPTR, QEvent::DeferredDelete);
+                remaining = timer.remainingTime();
+                if (remaining <= 0)
+                {
+                    break;
+                }
+                usleep(qMin(100, remaining) * 1000);
+                remaining = timer.remainingTime();
+            } while (remaining > 0);
+        }
+    }
     return result;
 }
 
--- a/src/modules/Extensions/YouTube.cpp
+++ b/src/modules/Extensions/YouTube.cpp
@@ -44,7 +44,6 @@
 #include <QLabel>
 #include <QMenu>
 #include <QUrl>
-
 #include <QDateTime>
 #include <QDeadlineTimer>
 
@@ -1307,8 +1306,12 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
             if (format.contains(QStringLiteral("available_at")))
             {
                 auto available_at = format[QStringLiteral("available_at")].toDouble() * 1000.0;
-                qCDebug(youtube) << "url for format" << format["format_id"].toString() << "will be available at"
-                    << QDateTime::fromMSecsSinceEpoch(available_at) << "so in" << (available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0 << "s";
+                double dt = (available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0;
+                if (dt < 0)
+                {
+                    qCDebug(youtube) << "url for format" << format["format_id"].toString() << "will be available at"
+                        << QDateTime::fromMSecsSinceEpoch(available_at) << "so in" << dt << "s";
+                }
                 if (available_at > latest_available_at)
                 {
                     latest_available_at = available_at;
@@ -1456,16 +1459,17 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
     {
         // The number of whole seconds until the "slowest" stream is promised to be available, plus 1
         // This interval can be <=0 if the stream is already available!
-        const auto dt = int( (latest_available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0) + 1;
-        if (dt > 0) {
-            QMPlay2Core.logInfo(QString::asprintf("Waiting for %ds as required by YouTube", dt));
+        const auto dt = static_cast<int>((latest_available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0) + 1;
+        if (dt > 0 && !youTubeDL.isAborted() )
+        {
+            const auto waitMsg = QString::asprintf("Waiting for %ds as required by YouTube", dt);
+            qCInfo(youtube) << waitMsg;
+            emit QMPlay2Core.statusBarMessage(waitMsg, dt * 1000);
             // do an active wait; a less active clone of QTest::qWait():
             auto remaining = dt * 1000.0;
             QDeadlineTimer timer(remaining, Qt::PreciseTimer);
             do
             {
-                QCoreApplication::processEvents(QEventLoop::AllEvents, remaining);
-                QCoreApplication::sendPostedEvents(nullptr, QEvent::DeferredDelete);
                 remaining = timer.remainingTime();
                 if (remaining <= 0)
                 {
@@ -1473,7 +1477,10 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
                 }
                 Functions::s_wait(qMin(100.0, remaining) / 1000.0);
                 remaining = timer.remainingTime();
-            } while (remaining > 0);
+            } while (remaining > 0 && !youTubeDL.isAborted());
+            if (remaining > 0) {
+                QMPlay2Core.logInfo(QStringLiteral("Stopped"));
+            }
         }
     }
     return result;


src/modules/Extensions/YouTube.cpp
```diff
@@ -1436,6 +1452,28 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
 
     result += o["description"].toString();
 
+    if (latest_available_at > 0)
+    {
+        const auto dt = int( (latest_available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0 + 0.5);
```
zaps166: why divide by 1000 and cast to `int` and later multipy by 1000 as `int` - no sense + integer overflow possible

src/modules/Extensions/YouTube.cpp
```diff
@@ -1436,6 +1452,28 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
...
+        if (dt > 0) {
+            QMPlay2Core.logInfo(QString::asprintf("Waiting for %ds as required by YouTube", dt));
+            // do an active wait; a less active clone of QTest::qWait():
+            auto remaining = dt * 1000;
+            QDeadlineTimer timer(remaining, Qt::PreciseTimer);
+            do
+            {
+                QCoreApplication::processEvents(QEventLoop::AllEvents, remaining);
+                QCoreApplication::sendPostedEvents(Q_NULLPTR, QEvent::DeferredDelete);
```
zaps166: `nullptr`

src/modules/Extensions/YouTube.cpp
```diff
@@ -1436,6 +1452,28 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
...
+                remaining = timer.remainingTime();
+                if (remaining <= 0)
+                {
+                    break;
+                }
+                usleep(qMin(100, remaining) * 1000);
```
zaps166: `Functions::s_wait()`

src/modules/Extensions/YouTube.cpp
```diff
@@ -44,6 +44,10 @@
 #include <QMenu>
 #include <QUrl>
 
+#include <QDateTime>
+#include <QDeadlineTimer>
+#include <unistd.h>
```
zaps166: Not needed

src/modules/Extensions/YouTube.cpp
```diff
@@ -1436,6 +1452,28 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
 
     result += o["description"].toString();
 
+    if (latest_available_at > 0)
```
zaps166, line 1458: How long it takes? Seconds/minutes/hours?

Does it properly abort on "stop" button or app close?

Is there any path which can call it in GUI thread (sorry - don't remember 😅)? If yes - it can't be like this.

src/modules/Extensions/YouTube.cpp
```diff
@@ -1298,6 +1304,16 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
                 urlLanguages[url] = format[QStringLiteral("language")].toString();
                 urlNotes[url] = note;
             }
+            if (format.contains("available_at"))
+            {
+                auto available_at = format["available_at"].toDouble() * 1000.0;
```
zaps166: Use `QStringLiteral` everywhere - be consistent.

RJVB:
On Saturday July 18 2026 15:14:58 Błażej Szczygieł wrote:

>why divide by 1000 and cast to `int` and later multipy by 1000 as `int` - no sense + integer overflow possible

Goal is to convert (rounding up) to whole seconds. I see that in fact I should probably  

Initially I just used `sleep(3)` but the simple, non-blocking wait function I picked up afterwards works with microseconds.

You're not wrong about the possibility of overflow, but that's not very likely to happen as long as `currentMSecsSinceEpoch()` returns a sensical value. But I now use a double inside the timer, since that's what s_wait() expects anyway.

BTW, I've tried to check for differences between this polite timer, and a simple, blocking call to usleep(), and couldn't find any evidence that the GUI freezes when I use the blocking timer.



>How long it takes? Seconds/minutes/hours?

I think the longest I've seen is 7 seconds. YouTube have no incentive to make real users wait for noticeable amounts of time.

>Does it properly abort on "stop" button or app close?

Bummer, apparently it does not. I probably have to check for some state variable, or do I need to set up a slot that gets connected to some signal in order to know if I need to interrupt the timer?
(I was expecting some function in QMPlay2Core to help here, but there isn't even an "isPlaying()" function?!)

>Is there any path which can call it in GUI thread (sorry - don't remember 😅)? If yes - it can't be like this.

As I said, the reason I use a timer (based on QTest::wait())` is to avoid freezing the GUI.
GetYouTubeVideo() is called only from YouTube::convertAddress(), but how would I know if this can be called from the main thread?

What's supposed to happen if you quit the application while it's processing the paste of a YT video into the playlist?

src/modules/Extensions/YouTube.cpp
RJVB: I suppose you meant the unistd.h header only.

src/modules/Extensions/YouTube.cpp
```diff
@@ -1436,6 +1452,30 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
 
     result += o["description"].toString();
 
+    if (latest_available_at > 0)
+    {
+        // The number of whole seconds until the "slowest" stream is promised to be available, plus 1
+        // This interval can be <=0 if the stream is already available!
+        const auto dt = int( (latest_available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0) + 1;
+        if (dt > 0) {
```
zaps166: Bracket to the newline.

src/modules/Extensions/YouTube.cpp
```diff
@@ -44,6 +45,9 @@
 #include <QMenu>
 #include <QUrl>
```
zaps166: Unneeded empty line

src/modules/Extensions/YouTube.cpp
```diff
@@ -1436,6 +1452,30 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
 
     result += o["description"].toString();
 
+    if (latest_available_at > 0)
+    {
+        // The number of whole seconds until the "slowest" stream is promised to be available, plus 1
+        // This interval can be <=0 if the stream is already available!
+        const auto dt = int( (latest_available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0) + 1;
```
zaps166: Hmm, if the stream is already available, should we wait or not? Also `static_cast<int>((latest_available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0) + 1)`.

> This interval can be <=0 if the stream is already available!

Should we wait in this case?

src/modules/Extensions/YouTube.cpp
```diff
@@ -1298,6 +1304,16 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
                 urlLanguages[url] = format[QStringLiteral("language")].toString();
                 urlNotes[url] = note;
             }
+            if (format.contains(QStringLiteral("available_at")))
+            {
+                auto available_at = format[QStringLiteral("available_at")].toDouble() * 1000.0;
+                qCDebug(youtube) << "url for format" << format["format_id"].toString() << "will be available at"
```
zaps166: Too much logs per every video. Put logs only if you need to wait (you did it already below).

src/modules/Extensions/YouTube.cpp
```diff
@@ -1436,6 +1452,30 @@ QStringList YouTube::getYouTubeVideo(const QString &param, const QString &url, I
 
     result += o["description"].toString();
 
+    if (latest_available_at > 0)
+    {
+        // The number of whole seconds until the "slowest" stream is promised to be available, plus 1
+        // This interval can be <=0 if the stream is already available!
+        const auto dt = int( (latest_available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0) + 1;
+        if (dt > 0) {
+            QMPlay2Core.logInfo(QString::asprintf("Waiting for %ds as required by YouTube", dt));
```
zaps166: I think it should go to stderr (`qCDebug`) + status bar as `statusBarMessage()`.
Well, logs in QMPlay2 are still bad.

zaps166 (commented):
Recently I noticed forbidden error on YouTube, so I can test it. Unfortunately, it's not fully working:

```
Extensions/YouTube: url for format "139" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "249" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "140" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "251" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "160" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "278" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "394" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "133" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "242" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "395" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "134" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "18" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "243" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "396" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "135" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "244" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "397" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "298" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "302" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "398" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "299" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "303" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
Extensions/YouTube: url for format "399" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
[https @ 0x7f9adc74b7c0] HTTP error 403 Forbidden
[https @ 0x7f9adc764840] HTTP error 403 Forbidden
```

RJVB:
>@zaps166 commented on this pull request.
>
>Recently I noticed forbidden error on YouTube, so I can test it. Unfortunately, it's not fully working:
>Extensions/YouTube: url for format "399" will be available at QDateTime(2026-07-25 12:24:32.000 CEST Qt::LocalTime) so in -0.432 s
>[https @ 0x7f9adc74b7c0] HTTP error 403 Forbidden
>[https @ 0x7f9adc764840] HTTP error 403 Forbidden

Oh damn. What's the link to that video? Does yt-dlp succeed?

> Also `static_cast<int>((latest_available_at - QDateTime::currentMSecsSinceEpoch()) / 1000.0) + 1)`.
Is there a difference with a regular, readable int() cast??


>> This interval can be <=0 if the stream is already available!
>
>Should we wait in this case?

I would hope not, because how long should one wait in that case? Take the absolute value of the dt and round that up?
I have not yet had 403 errors with this patch in place, but I think we'll have to look at the yt-dlp code (possibly make it print info like above) to see what it does. But as far as I can say now, there are times when it doesn't say that it will wait.

>Too much logs per every video. Put logs only if you need to wait (you did it already below).

OK for once we have this sorted out, but for now (and given your experience above) think it should at least log something for dt<=0.
 

>I think it should go to stderr (`qCDebug`) + status bar as `statusBarMessage()`.

It's an info message, so probably rather qCInfo or qCWarning, no?

>Well, logs in QMPlay2 are still bad.

No comment ;)

zaps166:
> Oh damn. What's the link to that video? Does yt-dlp succeed?

Video generally works, but fails once per few runs. It's not just one video (try https://www.youtube.com/watch?v=LXb3EKWsInQ). Same experience as without the patch.

```
Extensions/YouTube: url for format "139" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "249" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "140" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "251" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "160" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "278" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "330" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "694" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "133" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "242" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "331" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "695" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "134" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "18" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "243" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "332" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "696" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "135" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "244" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "333" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "697" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "298" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.634 s
Extensions/YouTube: url for format "302" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.635 s
Extensions/YouTube: url for format "334" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.635 s
Extensions/YouTube: url for format "698" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.635 s
Extensions/YouTube: url for format "299" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.635 s
Extensions/YouTube: url for format "303" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.635 s
Extensions/YouTube: url for format "335" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.635 s
Extensions/YouTube: url for format "699" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.635 s
Extensions/YouTube: url for format "308" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.635 s
Extensions/YouTube: url for format "336" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.635 s
Extensions/YouTube: url for format "700" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.635 s
Extensions/YouTube: url for format "315" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.635 s
Extensions/YouTube: url for format "337" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.635 s
Extensions/YouTube: url for format "701" will be available at QDateTime(2026-07-25 14:39:06.000 CEST Qt::LocalTime) so in -0.635 s
[25 Jul 2026 14:39:06.635] Waiting for 1s as required by YouTube
[https @ 0x7f67940ddc80] HTTP error 403 Forbidden
[https @ 0x7f6794000fc0] HTTP error 403 Forbidden
```

> Is there a difference with a regular, readable int() cast??

No difference, only syntax :sweat_smile:

RJVB:
On Saturday July 25 2026 05:42:04 Błażej Szczygieł wrote:

>Video generally works, but fails once per few runs. It's not just one video (try https://www.youtube.com/watch?v=LXb3EKWsInQ). Same experience as without the patch.

Hmmm, I haven't managed to get it to fail for me, but I get a wait time of 4-5 seconds each time.

>[25 Jul 2026 14:39:06.635] Waiting for 1s as required by YouTube

I only see negative wait times, so did you try my suggestion to use the absolute value of dt (rounded up)? If so, maybe increase the duration?

I wonder if this whole mechanism isn't a means to prevent too frequent accesses from a single client to a given video. At first I thought that part of the explanation is that they can't keep all formats of all videos available, but it doesn't make sense that you get the same wait time every time (or else it's because we don't play the entire video?). We may need to impose a minimum wait time (maybe even with an adaptive mechanism because there's clearly an effect of the location from which we connect).

zaps166:
Now hard coded 5 seconds (`Waiting for 5s as required by YouTube`) and still the same issue (once per few stop/play):
```
[https @ 0x7f2cfc009a40] HTTP error 403 Forbidden
[https @ 0x7f2cfc0070c0] HTTP error 403 Forbidden
```

---

Also you have a loop which doesn't check for `youTubeDL.isAborted()` - it should check for this instead of using event loop (it's running in thread, so the loop must abort if another thread asks for stop, but it only waits in separate thread, so no need for event loop i guess).

RJVB:
>zaps166 left a comment (zaps166/QMPlay2#1028)

>Now hard coded 5 seconds (`Waiting for 5s as required by YouTube`) and still the same issue (once per few stop/play):

So there's really no difference in the incidence rate with or without the patch?

What about yt-dlp, does it run into the same issue on this video?

FWIW, the negative dt values are <-2s for me.


>Also you have a loop which doesn't check for `youTubeDL.isAborted()`

Agreed on the probably no need for event checking here, but is youTubeDL running at this point? As far as I can tell, the wait currently comes after the call to youTubeDL.exec(), while some other thread is still waiting for the results of that.

Shouldn't there (also) be a call to isAborted() before doing all the json parsing work?

RJVB:
>@zaps166 commented on this pull request.

>Well, logs in QMPlay2 are still bad.

One thing you probably should do is to rename the Qt logging categories to something closer to the usual reverse URI domain notation. For instance, a line `org.qmplay2.extensions.youtube.debug=true` in qtlogging.ini would be easier to understand than `Extensions/YouTube.debug=true` ;)

RJVB:
René J.V. Bertin wrote on 20260725::16:35:15 re: "Re: [zaps166/QMPlay2] Try not to access YT streams before they're available (PR #1028)"

>FWIW, the negative dt values are <-2s for me.

Systematically the case for formats 92-94 and 300, 301 for me (available only when I use my Firefox cookies). And when I tell yt-dlp to download those it doesn't impose a wait.

zaps166:
> FWIW, the negative dt values are <-2s for me.

Maybe you have much slower machine. When I limit my CPU to 500 MHz, I also have ~-2s. It looks like time obtained by yt-dlp and time compared in QMPlay2 - ~2 seconds elapsed between point 1 and 2, so no need to wait at all. Anyway waiting doesn't help, I still can reproduce forbidden.

Maybe you should check for cookies, referer, user agent (it's QMPlay2 by default, maybe youtube randomly rejects it?).

RJVB:
On Thursday July 30 2026 16:02:05 Błażej Szczygieł wrote:

>> FWIW, the negative dt values are <-2s for me.
>
>Maybe you have much slower machine. When I limit my CPU to 500 MHz,

Probably much slower than yours, but not that much. It's an i7-6500U @2.50GHz, which should be fast enough (and I get the same wait times on a 2nd gen i7 @2.7Ghz that is noticeably slower).

Who knows what black magic that javascript challenge invokes that could introduce an effect of CPU speed on the delay with which videos are made available. And who knows what information is obtained from the browser cookies. I still think it's more likely that the difference in what we observe is linked to our geographic locations.

>Maybe you should check for cookies, referer, user agent (it's QMPlay2 by default, maybe youtube randomly rejects it?).

I don't change anything to the referer or user agent. I do know that the use of firefox's cookies makes a difference in the stream formats that yt-dlp returns, but the rejection occurs afterwards. 

In case you missed it: after a rejection I can reconstruct the list formula you use to open a video,audio pair directly with FFmpeg from the log. I have not yet managed to do this fast enough (= it always takes me a couple of seconds) and I have never had a rejection.

It's already a bit of a miracle that you can obtain stream URLs for formats that are available only if you have the right cookies set, and then access those streams without offering those same cookies.

Isn't there anything you can learn from the yt-dlp sources? They must have started to respect those `available_at` times for a reason (if they always did then only the notification about it is new, OR I simply never encountered videos for which a wait was required).

zaps166:
So what `available_at` means? I think it returns time of my PC (in UTC) when temporary URL was generated.

Maybe my occasional 403 reason is different than yours.

> I don't change anything to the referer or user agent. I do know that the use of firefox's cookies makes a difference in the stream formats that yt-dlp returns, but the rejection occurs afterwards.

I mean that you may try to use standard user agent instead of QMPlay2 :smile: and check results.

RJVB:
On Friday July 31 2026 11:49:54 Błażej Szczygieł wrote:

>So what `available_at` means? I think it returns time of my PC (in UTC) when temporary URL was generated.

I think it literally means "this stream will be available at UTC=x". Of course that should also be the time of your PC because evidently it is in perfect sync with the wall clock in our universe ;)

>Maybe my occasional 403 reason is different than yours.

That's what I was hinting at. There can be other reasons why access is forbidden even if you wait the amount of time you're supposed to wait.
One evident possibility: too frequent connection attempts (since you said it happens only "once in a while" ... that kind of suggests you're trying again and again).

>I mean that you may try to use standard user agent instead of QMPlay2 :smile: and check results.

Remind me, that would be if I run into rejections, right?

zaps166:
There are more and more 403 errors, also in M3U8 HLS live streams (also in mpv and ffmpeg) - live stream plays for less than a minute and freezes with 403 errors.

RJVB:
On Tuesday August 18 2026 03:53:19 Błażej Szczygieł wrote:

>There are more and more 403 errors, also in M3U8 HLS live streams (also in mpv and ffmpeg) - live stream plays for less than a minute and freezes with 403 errors.


Yes, and "the page needs to be reloaded" errors start popping up all of a sudden yesterday evening, even on videos I had justed looked at before.

See https://github.com/yt-dlp/yt-dlp/issues/17389, which contains a workaround.

zaps166:
:thinking: I don't use cookies from browser

RJVB:
On Tuesday August 18 2026 04:11:48 Błażej Szczygieł wrote:
>zaps166 left a comment (zaps166/QMPlay2#1028)
>
>:thinking: I don't use cookies from browser

From what I understand and see, that only makes matters worse. See the last comments in that yt-dlp ticket.

zaps166:
I see some posts there, but skimmed only. I tried nightly yt-dlp - same issue - downloads 39 seconds of live stream and then 403.

```
./yt-dlp "https://www.youtube.com/watch?v=..."
...
[https @ 0x55aae5d93fc0] HTTP error 403 Forbidden
[in#0/hls @ 0x55aae5284dc0] keepalive request failed for 'https://.../seg.ts' with error: 'Server returned 403 Forbidden (access denied)' when opening url, retrying with new connection
```

---

However URL extracted using "Video Download Helper" works properly :thinking:

RJVB:
On Tuesday August 18 2026 05:08:30 Błażej Szczygieł wrote:
>zaps166 left a comment (zaps166/QMPlay2#1028)
>
>I see some posts there, but skimmed only. I tried nightly yt-dlp - same issue - downloads 39 seconds of live stream and then 403.

I saw something about that on the ticket too.

>
>However URL extracted using "Video Download Helper" works properly :thinking: 

Heh, I watch a lot of probably not very legal videos by storing the URLs extracted with that extension O:-)

You can hopefully just wait until that livestream is finished and then watch it "in the replay". I'm doing that right now with https://www.youtube.com/live/2DLSWibujiY (almost 5 minutes in).

I also got confirmation that not waiting until the `available_at` time is likely to throw 403 errors, and that update channel switches are permanent.

So a priori, anyone bitten by the current situation could just do a manual switch (e.g. via the extra arguments thingy). It would help if you'd add a 24h update timer to the update-on-startup, or an "update now" button, I've been meaning to bring that up a while now. ;)

zaps166:
> I also got confirmation that not waiting until the `available_at` time is likely to throw 403 errors

Ok, but I want to test it; so far everything throws 403 xD

> and that update channel switches are permanent.

Understandable if the downloaded binary is from nightly channel, it'll download updates from that channel, too.

> It would help if you'd add a 24h update timer to the update-on-startup, or an "update now" button, I've been meaning to bring that up a while now. ;)

Instead of restarting QMPlay2? 24h timetout to re-check for updates seems to be good idea.

> You can hopefully just wait until that livestream is finished and then watch it "in the replay".

No if it's permanent live stream (TV).

RJVB:
On Tuesday August 18 2026 05:47:55 Błażej Szczygieł wrote:
>zaps166 left a comment (zaps166/QMPlay2#1028)
>
>> I also got confirmation that not waiting until the `available_at` time is likely to throw 403 errors
>
>Ok, but I want to test it; so far everything throws 403 xD

I think the nightly was just updated (2026.08.18.122307). I'm not seeing any 403s anymore, but if everything throws that error then you should probably be able to reproduce it when downloading with yt-dlp too, no?

In that case the additional output with `-vU` may contain some useful information aboud something different in your region ...

zaps166:
It's working so far :smile:

